PR #2, opened 2017-12-05
Use recommended way to specify Autoprefixer’ browsers

ai:
@devongovett Wow. Amazing project! Also focusing on PostCSS is neat =^_^=.

Small fix for `browsers` option. We don’t recommend to use them, since not only Autoprefixer need target browsers. ESLint/Stylelint plugins, cssnano, cssnext, next `babel-preset-env` 2.0 could read browsers from one config file: [.browserslistrc](https://github.com/ai/browserslist).

We recommend only this way. Seems like you wrote this option to show options syntax. So I changed the option to `grid`.

### DIFF
--- a/src/docs/transforms.md
+++ b/src/docs/transforms.md
@@ -39,13 +39,20 @@ Then, create a `.postcssrc`:
   "modules": true,
   "plugins": {
     "autoprefixer": {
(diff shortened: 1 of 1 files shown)


devongovett:
Thanks!

PR #14, opened 2017-12-06
Prettier

jamesmcallister:
Added Prettier and formated current files


DanielRuf:
Better ignore the package-lock file and rebase on the current master.

PR #16, opened 2017-12-06
Async functions producing an error "ReferenceError: regeneratorRuntim…

albinotonnina:
…e is not defined"

Maybe this could save some time to people trying dynamic import with async/await syntax?


src/docs/code_splitting.md
```diff
@@ -35,3 +35,28 @@ async function renderPage(page) {
   return render();
 }
 ```
+
+If you decide to go for the async/await syntax remember to load babel-polyfill (tranform-runtime in case of libraries) and to relocate the code containing the async function into a separate module
+
+Refer to this issue: [babel#5085](https://github.com/babel/babel/issues/5085#issuecomment-277544677)
```
thejameskyle: I don't think we need to link to that issue

src/docs/code_splitting.md
```diff
@@ -35,3 +35,28 @@ async function renderPage(page) {
...
+
+```bash
+yarn add babel-polyfill
+yarn add babel-preset-stage-0
```
thejameskyle: We should not recommend using `stage-0`, very few people should ever use that. `babel-preset-env` covers async functions anyways.

PR #22, opened 2017-12-07
Typo in Plugins

lahmatiy:


devongovett:
the dot is not required.

lahmatiy:
Well, on https://parceljs.org/asset_types.html

```js
bundler.addAssetType('.ext', require.resolve('./MyAsset'));
```

on https://parceljs.org/plugins.html
```
bundler.addAssetType('ext', require.resolve('./MyAsset'));
```

Both examples are correct? I think it should be noted then or fixed.

PR #29, opened 2017-12-08
Update dynamic import docs to clarify default imports

thejameskyle:
To be merged with: https://github.com/parcel-bundler/parcel/pull/78


devongovett:
@kingdaro want to make a follow up PR?

PR #35, opened 2017-12-11
Korean v.1.0

march23hare:
Translated HTML files and MD docs are placed in src/i18n/kor.
Kor directory's structure same as src directory's.
I hope that it is useful.


brandon93s:
This is great!

It may require a custom build script vs the existing npm scripts, but we should be able to map this to `/ko/index.html` (and others in the future) to be served by serve. 

@march23hare are you interested in taking a look at the build process ([markdown-styles](https://github.com/mixu/markdown-styles)) and deployment to see if a build pipeline can be setup? Glad to assist on this front!

march23hare:
Thank you again @brandon93s.
I'm going to begin studying about markdown-styles and Zeit Now.

PR #40, opened 2017-12-15
Document edge case with HMR for non-JS

real34:
... I think it could help further troubleshooting!

Feel free to reword it as you'd prefer of course.

The basic thing is that without any JS script included in the HTML Parcel will not inject its script containing the HMRClient initialization. I think it can be considered as a feature, but if you think it would be better to "fix" this behavior to inject a new script tag do not hesitate to close this PR and create an issue in the main repo.

### DIFF
--- a/src/i18n/en/docs/hmr.md
+++ b/src/i18n/en/docs/hmr.md
@@ -17,3 +17,5 @@ if (module.hot) {
   });
 }
 ```
+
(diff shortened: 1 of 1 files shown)


devongovett:
This is a bug, and is being worked on here: https://github.com/parcel-bundler/parcel/pull/128

PR #50, opened 2017-12-16
Fix language selection using subdomains on localhost

fathyb:
This let contributors test the language they are translating using query strings (eg. `localhost:5000/?locale=ru`).

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -6,10 +6,9 @@ const languages = fs.readdirSync(__dirname + '/../dist');
 
 app.use(function (req, res, next) {
   // autodetect language
-  let lang = req.subdomains[0] || req.acceptsLanguages(...languages) || 'en';
-  if (!languages.includes(lang)) {
-    lang = 'en';
-  }
+  let lang = [req.query.locale, req.subdomains[0], req.acceptsLanguages(...languages), 'en'].find(function(lang) {
+    return languages.includes(lang);
+  });
 
   req.url = '/' + lang + req.url;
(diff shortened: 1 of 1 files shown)


src/server.js
```diff
@@ -5,8 +5,11 @@ const app = express();
 const languages = fs.readdirSync(__dirname + '/../dist');
 
 app.use(function (req, res, next) {
+  // this expression looks for the lang subdomain, and makes an exception for localhost
+  const subdomain = req.hostname.match(/^([^\.]+)\.((localhost)|([^\.]+\.[^\.]+))/)
```
davidnagli: @fathyb Don’t you think it would be more efficient to add a `--dev` CLI option to `server.js`? I feel like running a complex regex on every single request might be expensive.

fathyb: Yep I agree, and this is hacky and makes the code more hard to understand. The goal is just to make it easier to test a specific language using the server locally, what do you think of just passing the language to the server, like `yarn start ko`?

davidnagli: I think the easiest way is to pass in an environment variable called `dev`, and just check if it’s true/false. 

Then, we can make an npm script that allows us to use it like so: `yarn dev`

brandon93s: It may be easier just to accept a query string locale. This could be left for both dev and production.

`http://someurl.org/page.html?locale=ru`

And then this could just be added to the cascading checks.

davidnagli: That works too

fathyb: Pushed 🙂

PR #55, opened 2017-12-17
Fix Ru translation and mistypes

koddr:


davidnagli:
@devongovett Before merging, note that the words “copyright” and your name are now written in Russian. While it looks better this way, I’m not sure if it has any legal implications.

Just keep that in mind…

PR #58, opened 2017-12-20
Optimize images with imageoptim lossy compression @ 70%

ryanworl:
I optimized the PNG images with [ImageOptim](https://imageoptim.com/mac), which uses lossy compression to reduce PNG file sizes, which I saw where a problem on the homepage.

I used the 70% quality level.

I don't notice any difference in quality.

This reduces the overall page weight by over 1 megabyte on browsers that request the @3x images.


devongovett:
Awesome, thanks!

PR #65, opened 2017-12-24
adding doc session about how to change language

lai32290:
Adding doc session about how to change language

README file visual result


lai32290:
I spent some time reading `src/server.js` file to figure out about how to change language, so I added it to `README` file, trying to make it easier for next one.
I'm not sure it's worth or not, feel free to change it or just close the PR.

PR #66, opened 2017-12-24
adding config for server to accept locale as arguments

lai32290:
Adding configuration to `src/server.js` to accept locale like: 
* `node ./src/server.js -l ko`
* `node ./src/server.js --locale ko`
* `node ./src/server.js --locale=ko`
* `node ./src/server.js --locale='ko'`

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "express": "^4.16.2"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


lai32290:
I tried to access by `en.localhost` but it's not working, I'm using a fake domain now, I'll close the PR, sorry about that

PR #69, opened 2017-12-27
Add node version 6+ requirement in english and portuguese

lrn2prgrm:
As discussed in the Slack this information need to get added. I attach my proposal in this two languages.

### DIFF
--- a/src/i18n/en/docs/getting_started.md
+++ b/src/i18n/en/docs/getting_started.md
@@ -2,7 +2,7 @@
 
 Parcel is a web application bundler, differentiated by its developer experience. It offers blazing fast performance utilizing multicore processing, and requires zero configuration.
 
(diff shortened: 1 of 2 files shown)


brandon93s:
I missed the conversations in Slack, but Node 6 support is in progress: https://github.com/parcel-bundler/parcel/pull/253. I believe our intent is to continue to support Node 6.

lrn2prgrm:
Should I edit to say 6+ then? That information is still missing

lrn2prgrm:
@brandon93s now says node 6+. I think it is still important to say it on getting started.

DeMoorJasper:
@lrn2prgrm I don't really think anyone uses a version below Node 6 and it will pop an error on installation anyway?

brandon93s:
There is still plenty of usage on Node < 6 ([stats](https://nodejs.org/metrics/)), but since npm & yarn will prompt on failure we'll close this out.

PR #81, opened 2018-01-05
Spanish translation

kidandcat:
Full Spanish translation.

### DIFF
--- a/src/i18n/en/index.html
+++ b/src/i18n/en/index.html
@@ -39,6 +39,7 @@
           <li><a href="https://pl.parceljs.org">🇵🇱</a></li>
           <li><a href="https://ru.parceljs.org">🇷🇺</a></li>
           <li><a href="https://zh.parceljs.org">🇨🇳</a></li>
+          <li><a href="https://es.parceljs.org">🇪🇸</a></li>
         </ul>
       </div>
       <a href="getting_started.html">Documentation</a>
--- a/src/i18n/en/layout/page.html
+++ b/src/i18n/en/layout/page.html
@@ -30,6 +30,7 @@
           <li><a href="https://ko.parceljs.org">🇰🇷</a></li>
           <li><a href="https://ru.parceljs.org">🇷🇺</a></li>
           <li><a href="https://zh.parceljs.org">🇨🇳</a></li>
+          <li><a href="https://es.parceljs.org">🇪🇸</a></li>
         </ul>
       </div>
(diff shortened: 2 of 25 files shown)


kidandcat:
It looks like travis CI is stuck or broken as it couldn't update the ubuntu repositories and did not finish the npm install (looks like it has no internet connection).

gnapse:
Full disclosure, I'm part of the team that authored the other PR for Spanish translation (#87). 

Among the differences I noticed with this one, is that this one is using the formal way to address the reader, whereas we preferred the informal and more casual way of doing so.

Also I noticed that some of the section titles were left untranslated here ([here](https://github.com/kidandcat/website-1/blob/1fcc1f959ee371b3b298004ef9f04c758b882b71/src/i18n/es/docs/assets.md), [here](https://github.com/kidandcat/website-1/blob/1fcc1f959ee371b3b298004ef9f04c758b882b71/src/i18n/es/docs/code_splitting.md) and [here](https://github.com/kidandcat/website-1/blob/1fcc1f959ee371b3b298004ef9f04c758b882b71/src/i18n/es/docs/getting_started.md) at least).

gnapse:
Yup, you've probably got a point in there with Code Splitting and/or Hot Module Replacement, but I'd argue that perhaps a middle ground is best. Because it's hard to draw the line in what terms are best understood in English and which ones are translatable. I'd say that the title should be in Spanish, and somewhere in the beginning of the documentation of each chapter, a reference to the English term should be made, or maybe in the title it self (i.e. "Reemplazo de Módulos en Caliente (Hot Module Replacement)") (which we did not do in our translation either, we went the way of a more fully Spanish experience).

The [assets](https://github.com/kidandcat/website-1/blob/1fcc1f959ee371b3b298004ef9f04c758b882b71/src/i18n/es/docs/assets.md) file is one where I'd argue that it has to be translated to. I know the word "Assets" is probably more recognizable even by Spanish-speaking developers than "Recursos" o "Activos", but again, what's the point of translating the site if we're going to leave all this in English? A quick survey over all the other translated Parcel websites already deployed tells me that only the Polish devs decided to leave a single module title in English (HMR), but all the rest were translated. I'd say we should strive for that or all, and put alongside the English equivalent for cases like these.

Frankly, at this point I'd say your PR should be accepted on the grounds that it came in first. I just pointed out these issues not to criticize and enhance our PRs odds of being merged instead, but because I think some of them maybe should be addressed before merging this. Or could also be fixed by the community afterwards with new PRs, if consensus is reached.

brandon93s:
`es` landed with #87

Issue #57, opened 2017-12-18
Translate parcel.org into Spanish


dgrizzla:
Hi, I haven't found any issue related to spanish docs, so I've decided to team up with @albizures to translate the site.

devongovett:
Awesome, let me know if you need anything!

lrn2prgrm:
@dgrizzla, I could colaborate. let me know if you want to assign some files to me so I can for your fork and send you some PRs. I will write to your email too.

clarksandholtz:
I'd love to help if there is still work to be done!

dgrizzla:
@clarksandholtz sounds great! There are some files that still need to be translated, check out the repository and hit me up with the files you're gonna change. Thanks in advance!

devongovett:
Is this PR related to the work going on here? https://github.com/parcel-bundler/website/pull/81

dgrizzla:
@devongovett not at all

devongovett:
Hmm, ok @kidandcat would be good if you could collaborate with @dgrizzla and others here.

devongovett:
And next time please do check the issues to see if someone is already working on it. That way we don't duplicate work.

kidandcat:
@devongovett Sorry, didn't even notice there was something going on.

kidandcat:
UP
This is going out of control. There are now 2 PR for a full Spanish translation, has it been any of you?

gnapse:
No, sorry, I too started working on this with a couple of colleagues. I mentioned it around, but did not check for issues like this one. So it's three efforts so far, and two of them out on PRs (none of which apparently are from the creators of this issue). Sorry!

devongovett:
Done!

PR #87, opened 2018-01-10
Spanish translation

gnapse:
Spanish translation for the website.


kidandcat:
@devongovett Could you accept this PR soon?

We want to spread ParcelJS to the spanish community. I can't understand why it has to go so long when there are PR's (and dedicated people) like this one fully translated and ready to be deployed.

gnapse:
Thanks @brandon93s! And thanks again to everyone who contributed, including @kidandcat and the folks on #57 that also started on a similar effort. Fixed on top of this work are welcome.

BTW @brandon93s, the website https://es.parceljs.org/ is not loading correctly. It appears to have some sort of issue with the certificate, probably not being associated still with the domain or something like that.

devongovett:
Awesome, thanks for your work on this! Sorry for my slow responses, I've been super busy.

kidandcat:
@devongovett Thanks for your time, and thanks @gnapse for the translation! let's make a great community around ParcelJS!

PR #101, opened 2018-01-24
Update index.html

rbrtmrtn:
Minor edit to add a hyphen to  "zero-configuration"


DeMoorJasper:
I think both are correct, i might be mistaken though

brandon93s:
Let's keep it as "zero configuration". We've consistently gone without the hyphen. A change here would necessitate across the GitHub org, repos, readmes and other localized pages.

PR #108, opened 2018-02-06
translate(zh-cn): translate safe write

QC-L:


src/i18n/zh/docs/hmr.md
```diff
@@ -1,6 +1,6 @@
-# 🔥 热模块替换(HMR)
+# 🔥 热模块重载(HMR)
```
dear-lizhihua: hmr 应该是模块热重载/模块热替换吧

PR #116, opened 2018-02-18
API Docs

### LINKED ISSUE #114: Document bundler events
e.g. `bundler.on("bundled", ...)`

> Seems reasonable to use the API for this I think. We should probably document it so people know they can do that though!

https://github.com/parcel-bundler/parcel/issues/800

DeMoorJasper:
Adds api docs for middleware, Bundle, events and Bundler

closes #114

### DIFF
--- a/src/i18n/en/docs/api.md
+++ b/src/i18n/en/docs/api.md
@@ -1,0 +1,136 @@
+# 📚 API
+
+## Bundler
+
+Instead of the cli you can also use the API to initialise a bundler, for more advanced use-cases (e.g. custom processing after every build).
+A watch example with every option explained:
+```Javascript
+const Bundler = require('parcel-bundler');
+const Path = require('path');
+
+// Entrypoint file location
+const file = Path.join(__dirname, './index.html');
+
+// Bundler options
+const options = {
+  production: false, // build or watch, defaults to process.env.NODE_ENV === 'production'
+  outDir: './dist', // The out directory to put the build files in, defaults to dist
+  outFile: 'index.html', // The name of the outputFile
+  publicUrl: './', // The url to server on, defaults to dist
+  watch: true, // whether to watch the files and rebuild them on change, defaults to !production
+  cache: true, // Enabled or disables caching, defaults to true
+  cacheDir: '.cache', // The directory cache gets put in, defaults to .cache
+  minify: false, // Minify files, defaults to minifying if production is true
+  target: 'browser', // browser/node/electron, defaults to 
+  https: false, // Server files over https or http, defaults to browser
+  logLevel: 3, // 3 = log everything, 2 = log warnings & errors, 1 = log errors
+  hmrPort: 0, // The port the hmr socket runs on, defaults to a random free port (0 in node.js resolves to a random free port)
+  sourceMaps: true, // Enable or disable sourcemaps, defaults to enabled (not supported in production builds yet)
+  hmrHostname: '', // A hostname for hot module reload, default to ''
+  detailedReport: false // Prints a detailed report of the bundles, assets, filesizes and times, defaults to false, reports are only printed if watch is disabled
+};
+
+// Initialises a bundler using the entrypoint location and options provided
+const bundler = new Bundler(file, options);
+
+// Run the bundler, this returns the main bundle
+// Use the events if you're using watch mode as this promise will only trigger once and not for every rebuild
+const bundle = await bundler.bundle();
+```
+
+### Events
+
+This is a list of all bundler events
+
+`bundled` gets called once parcel has succesfully finished bundling, it has the main [bundle](#bundle) as a parameter
+
+```Javascript
+const bundle = new Bundler(...);
+bundle.on('bundled', (bundle) => {
+  // bundle contains all assets and bundles, see documentation for details
+});
+```
+
+`buildEnd` gets called after each build, this also emits if an error accured
+
+```Javascript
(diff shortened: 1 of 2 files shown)


src/i18n/en/docs/api.md
```diff
@@ -0,0 +1,136 @@
...
+```Javascript
+const Bundler = require('parcel-bundler');
+const Path = require('path');
+
+// Entrypoint file location
+const file = Path.join(__dirname, './index.html');
+
+// Bundler options
+const options = {
+  production: false, // build or watch, defaults to process.env.NODE_ENV === 'production'
```
mischnic: isn't this production vs development as `watch` is a seperate key?

src/i18n/en/docs/api.md
```diff
@@ -0,0 +1,136 @@
...
+  outDir: './dist', // The out directory to put the build files in, defaults to dist
+  outFile: 'index.html', // The name of the outputFile
+  publicUrl: './', // The url to server on, defaults to dist
+  watch: true, // whether to watch the files and rebuild them on change, defaults to !production
+  cache: true, // Enabled or disables caching, defaults to true
+  cacheDir: '.cache', // The directory cache gets put in, defaults to .cache
+  minify: false, // Minify files, defaults to minifying if production is true
+  target: 'browser', // browser/node/electron, defaults to
```
mischnic: > defaults to *browser*

src/i18n/en/docs/api.md
```diff
@@ -0,0 +1,136 @@
...
+  production: false, // build or watch, defaults to process.env.NODE_ENV === 'production'
+  outDir: './dist', // The out directory to put the build files in, defaults to dist
+  outFile: 'index.html', // The name of the outputFile
+  publicUrl: './', // The url to server on, defaults to dist
+  watch: true, // whether to watch the files and rebuild them on change, defaults to !production
+  cache: true, // Enabled or disables caching, defaults to true
+  cacheDir: '.cache', // The directory cache gets put in, defaults to .cache
+  minify: false, // Minify files, defaults to minifying if production is true
+  target: 'browser', // browser/node/electron, defaults to 
+  https: false, // Server files over https or http, defaults to browser
```
mischnic: Here is the browser for the previous line

src/i18n/en/docs/api.md
```diff
@@ -0,0 +1,136 @@
...
+// Run the bundler, this returns the main bundle
+// Use the events if you're using watch mode as this promise will only trigger once and not for every rebuild
+const bundle = await bundler.bundle();
+```
+
+### Events
+
+This is a list of all bundler events
+
+`bundled` gets called once parcel has succesfully finished bundling, it has the main [bundle](#bundle) as a parameter
```
mischnic: > ..., the main bundle gets passed to the callback

or something like that

src/i18n/en/docs/api.md
```diff
@@ -0,0 +1,136 @@
...
+
+// Initialises a bundler using the entrypoint location and options provided
+const bundler = new Bundler(file, options);
+
+// Run the bundler, this returns the main bundle
+// Use the events if you're using watch mode as this promise will only trigger once and not for every rebuild
+const bundle = await bundler.bundle();
+```
+
+### Events
```
mischnic: maybe format the events as
- a
- list?

src/i18n/en/docs/api.md
```diff
@@ -0,0 +1,136 @@
...
+```Javascript
+const bundle = new Bundler(...);
+bundle.on('buildEnd', () => {
+
+});
+```
+
+### Bundle
+
+A `Bundle` is what parcel uses to bundle assets together, this also contains child and sibling bundles to be able to build a bundle tree.
```
mischnic: It's not clear what child and sibling bundles are. Child bundles get included in a parent bundle and sibling bundles are seperate files?

src/i18n/en/docs/api.md
```diff
@@ -0,0 +1,136 @@
...
+* `offsets`: A `Map<Asset, number(line number inside the bundle)>` of all the locations of the assets inside the bundle, used to generate accurate source maps
+
+#### Tree
+
+The `Bundle` contains `parentBundle`, `childBundles` and `siblingBundles`, all these properties together create a fast to iterate bundle tree.
+
+
+A very basic example of an asset tree and it's generated bundle Tree
+
+Asset tree:
```
mischnic: Is there a subsubsub-section-heading-level left for Asset/Bundle tree? Otherwise bold to make it more apparent.

src/i18n/en/docs/api.md
DeMoorJasper: Probably a good idea, have changed it

src/i18n/en/docs/api.md
DeMoorJasper: I hope it's more clear now, not sure if it needs a lot of explaining, as child, sibling and parent are pretty self-explanatory

src/i18n/en/docs/api.md
DeMoorJasper: There is, changed it

src/i18n/en/docs/api.md
DeMoorJasper: It can be used seperately but i guess `process.env.NODE_ENV` is probably the recommended way of setting this

devongovett:
Would be great to get this new page translated into other languages!

PR #133, opened 2018-03-28
Add Vue recipe

f-a-a:
Following from the v1.7.0 release. I just thought I could contribute a little. Hope it's helpful!

### DIFF
--- a/src/i18n/en/docs/recipes.md
+++ b/src/i18n/en/docs/recipes.md
@@ -81,3 +81,41 @@ Add Start script to `package.json`
   "start": "parcel index.html"
 }
 ```
+
+## Vue
+
+First we need to install the dependencies for Vue.
+
+```bash
+npm install --save vue
+npm install --save-dev parcel-bundler
+npm install --save-dev babel-preset-env
+npm install --save-dev babel-preset-vue
+```
+
+<sub>Or if you have the optional Yarn package manager installed</sub>
+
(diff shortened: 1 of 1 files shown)


src/i18n/en/docs/recipes.md
```diff
@@ -81,3 +81,41 @@ Add Start script to `package.json`
...
+yarn add --dev babel-preset-vue
+```
+
+Then make sure the following Babel config is present.
+
+```javascript
+// .babelrc
+{
+  "presets": ["env", "vue"]
+}
```
devongovett: This babel preset is probably not needed. env support is built-in to parcel automatically

devongovett: also what is babel-preset-vue for?

DeMoorJasper: It's for Vue JSX, so I think it shouldn't be necessary as we use vue-compiler-utils before running it through babel

devongovett: Oh, maybe we should just add auto JSX support for Vue like we have for React/Preact etc.

devongovett: I didn't realize Vue even used JSX

f-a-a:
@devongovett @DeMoorJasper I've updated the PR to remove the unneeded dependencies.

Issue #132, opened 2018-03-22
Document Parcel v1.7.0 features


davidnagli:
Parcel 1.7.0 is coming out soon, and it's gonna be a really huge update with tons of very exciting features and changes! 🚀 😃 

The only problem is... none of it is documented yet 😟 

I think we should all work together on getting the major changes documented in depth:
- [ ]  **New Resolver** (parcel-bundler/parcel#850) - Explain tilde/absolute paths, aliases, etc.
- [ ]  **Auto Install Feature**  (parcel-bundler/parcel#805) - There's not much to explain, but it's still new concept so we're gonna need to document what it and why it's useful
- [ ] **New File Naming Strategy**  (parcel-bundler/parcel#1025) - We should add something to the advanced documentation with that table that @jamiebuilds made in the PR comments

Lmk if I missed anything else that needs to be documented.

devongovett:
* GLSL assets
* New default public url (probably change the one here to something else https://parceljs.org/production.html#set-the-public-url-to-serve-on)
* Custom SSL certificate options

DeMoorJasper:
* Update https://parceljs.org/asset_types.html to the improved pipeline api

item4:
I think you missed about`.vue` file support and `--log-level`

wesleyboar:
[1.70 has been released](https://github.com/parcel-bundler/parcel/releases). I am very much looking forward to some documentation. Where should I ask for help until that documentation is available?

> I am trying to use `~/` for protject-root-relative paths—and I am failing…
> 
> ```
> /omg/absolute/path/scripts/src/site/site.js:21:28: Cannot resolve dependency '~/src/site/visual/is-logged-in.js' at '/omg/absolute/path/scripts/src/site/~/src/site/visual/is-logged-in.js'
> ```
> 
> The nearest `package.json` is at `/omg/absolute/path/scripts/package.json`.

TimNZ:
The docs have been for 1.7.0 for around a week.

A prominent version number on the docs pages would be helpful.

Devon blogged on 1.7.0.
https://medium.com/@devongovett/parcel-v1-7-0-9aac0c505837

A **Blogs** or **Releases** link on the landing page along with current version number would be useful

davidnagli:
@wesleyboar I think most things should be documented, but if you ever run into something you need help with you can always message us on the Parcel slack :)

PR #134, opened 2018-03-29
Update some docs to 1.7

DeMoorJasper:
* Update Asset to use new pipeline array return from generate function
* Create seperate section for cli commands and options
* Update React recipe to no longer need a .babelrc or babel dependency

Related #132


src/i18n/en/docs/asset_types.md
```diff
@@ -12,12 +12,12 @@ const {Asset} = require('parcel-bundler');
 class MyAsset extends Asset {
   type = 'foo'; // set the main output type.
 
-  parse(code) {
+  async parse(code) {
```
davidnagli: Did we make a breaking API change that makes this async? Doesn't that mean we broke existing plugins?

DeMoorJasper: It's not async I just thought i'd make more sense to change it as most parsers use localRequire and such making it async. No breaking changes.

davidnagli: Oh got you 👍

src/i18n/en/docs/recipes.md
```diff
@@ -10,8 +10,6 @@ First need to install the dependencies for React.
 npm install --save react
 npm install --save react-dom
```
davidnagli: Now that we have autoinstall we can get rid of these imports entirely :)

davidnagli: Funny enough that actually completely eliminates the need for the recipe... it would just be empty.

I guess we can just have only one thing under the recipe:
```
parcel src/index.js
```
and then just have a little thing below it explaining that this will automatically pull in all deps and config babel for you.

davidnagli:
Whoops didn't realize Devon already merged this. I'll open up a followup PR 😄

PR #139, opened 2018-04-07
ko: translate v1.7.0 things

item4:


item4:
@Hazealign how do you think about "빠른 모듈 교체"?

Hazealign:
@item4 That might be good. And "핫 모듈 교환" also will be a good alternative. (vuex's korean translation use this.)

DeMoorJasper:
Ready to merge?

PR #196, opened 2018-07-05
Add note for parcel's auto-install feature

DeMoorJasper:
Adds a note to hmr about auto-install

Related to #132

### DIFF
--- a/src/i18n/en/docs/hmr.md
+++ b/src/i18n/en/docs/hmr.md
@@ -18,6 +18,14 @@ if (module.hot) {
 }
 ```
 
+## Automagically installed dependencies
+
+Whenever Parcel comes across a dependency that fits the `node_module` pattern and can't find it, Parcel tries to install this dependency using `yarn` or `npm` depending on finding a `yarn.lock` file or not. This prevents the developer from having to exit parcel or having multiple terminal windows open.
+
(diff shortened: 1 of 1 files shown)


src/i18n/en/docs/hmr.md
```diff
@@ -18,6 +18,14 @@ if (module.hot) {
 }
 ```
 
+## Automagically installed dependencies
+
+Whenever Parcel comes across a dependency that fits the `node_module` pattern and can't find it, Parcel tries to install this dependency using `yarn` or `npm` depending on finding a `yarn.lock` file or not. This prevents the developer from having to exit parcel or having multiple terminal windows open.
+
+This only occurs in *development* (using [`serve`](./cli.html#serve) or [`watch`](./cli.html#watch)), in production (using [`build`](./cli.html#serve)) autoinstall is disabled to prevent unwanted side-effects on deployment.
+
+You can disable this feature using [`--no-autoinstall`](./cli.html#disable-autoinstall).
```
forresst: This link does not work, please replace by:

```markdown
You can disable this feature using [`--no-autoinstall`](cli.html#disable-autoinstall).
```

***In general, it would be necessary to remove `. \` for the 4 added links***

DeMoorJasper:
@forresst would be awesome if you could open up a PR

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

forresst:
Done: #199

PR #210, opened 2018-07-18
Update cssnano docs to accommodate v4.0.0 presets

heldr:


heldr:
based on https://github.com/parcel-bundler/parcel/issues/698#issuecomment-361633383

PR #224, opened 2018-08-17
Documentation for custom ssl certificate and key in API

hirasso:
Hope everything looks right – first PR ever 😅


hirasso:
Yey, great success! ;))

> Just slightly updated it to prevent confusion with duplicate https keys

You are right, it's much more clear like that.

PR #227, opened 2018-08-21
Add note about the hmr problem: "import first, create imported later"

MatejMazur:
[Related comment](https://github.com/parcel-bundler/parcel/issues/1915#issuecomment-414754759)

It's not the best PR :smile: but it would be great to somehow explain that if Parcel doesn't refresh after an error and you know that you repair this error then you should try to resave (Ctrl + S) some of the resolved files.

I'm fine if you find a better solution. :slightly_smiling_face:

### DIFF
--- a/src/i18n/en/docs/hmr.md
+++ b/src/i18n/en/docs/hmr.md
@@ -8,11 +8,11 @@ There are two methods to know about: `module.hot.accept` and `module.hot.dispose
 
 ```javascript
 if (module.hot) {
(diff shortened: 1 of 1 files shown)


DeMoorJasper:
This actually sounds more like a HMR bug than something we should mention in the docs, I originally thought you were gonna add a note for the always refresh HMR.

Not sure though, might be interesting to add a sidenote on what to do if it doesn't work.

PR #229, opened 2018-08-23
document module resolution strategies

### LINKED ISSUE #184: docs: add section on module resolution
I am currently playing around with Parcel on a new React app I created. Almost everything I do in JavaScript, I do in TypeScript. With that being said, I take advantage of Module Resolution. I know that absolute paths/imports were finalized in the 1.7.0 release and in this  [PR](https://github.com/parcel-bundler/parcel/pull/850), but I'd like to document this.

I read some comments and while I can respect different perspectives, I know that for larger projects I work on, it's important to use absolute imports or the imports themselves get too deep/convoluted. Trying out the solution in the PR mentioned above, I do have absolute imports working but my IDE (VS Code) is going crazy with errors. This doesn't affect the actual app running but can be distracting when I am trying to debug. I think a new section on module resolution / absolute imports with Parcel can help developer experience and adoption with Parcel. (This was the biggest reason I didn't try Parcel. I just learned that this feature existed the other day.)

I can't commit to having this PR in an hour or even a week. I'd like to fully understand how everything works so I can provide some stellar docs for other people. If this sounds good to project maintainers, please let me know! Thanks for all the hard work, ladies and gents.

jthegedus:
This is my initial attempt at resolving https://github.com/parcel-bundler/website/issues/184, https://github.com/parcel-bundler/website/issues/132 and documenting https://github.com/parcel-bundler/parcel/pull/850

I have a few questions which I believe should be answered and better described before merging:
- I would like to answer this question in the docs https://github.com/parcel-bundler/parcel/pull/850#issuecomment-372105317
- clarify terminology of:
	- project root: top-level package.json?
	- package root: package.json in node_module module? where does this start to resolve from?
- these comments https://github.com/parcel-bundler/parcel/pull/850/files#diff-f90b0caefa8028ff5ec8f2f2c2e6e39dR8 seem to suggest I may be missing some important information that should be documented

Closes #184

### DIFF
--- a/src/i18n/en/docs/module_resolution.md
+++ b/src/i18n/en/docs/module_resolution.md
@@ -1,6 +1,6 @@
 # :notebook_with_decorative_cover: Module Resolution
 
-Parcel (v1.7.0 and above) supports multiple module resolution strategies out of the box so you don't have to deal with endless `../../`.
+Parcel (v1.7.0 and above) supports multiple module resolution strategies out of the box so you don't have to deal with endless relative paths `../../`.
 
 ## Absolute Paths
 
@@ -33,7 +33,7 @@ This example aliases `react` to `preact` and some local custom module that is no
 
 Avoid using `~` in your aliases as it is already used by parcel as mentioned earlier.
 
-## Language Specific Conditions
+## Other Conditions
 
 ### JavaScript Named Exports
 
@@ -66,3 +66,7 @@ TypeScript will need to know about your use of the `~` module resolution or alia
       "~*": ["./src/*"]
     },
 ```
+
+### Monorepo Resolution
+
+TODO: answer this - https://github.com/parcel-bundler/parcel/pull/850#issuecomment-372105317
--- a/src/i18n/en/docs/module_resolution.md
+++ b/src/i18n/en/docs/module_resolution.md
@@ -44,10 +44,12 @@ Alias mappings apply to many asset types and does not specifically support mappi
 {
   "name": "some-package",
   "alias": {
-    "ipcRenderer": "./electron-ipc.js"
+    "ipcRenderer": "./electron-ipc.js" // specify file extension
   }
 }
 ```
+
+and re-export the named export within the aliased file:
 
 ```js
 // electron-ipc.js
--- a/src/i18n/en/docs/module_resolution.md
+++ b/src/i18n/en/docs/module_resolution.md
@@ -69,13 +69,14 @@ module.exports = require("electron").ipcRenderer;
 TypeScript will need to know about your use of the `~` module resolution or alias mappings. Please refer to the [TypeScript Module Resolution docs](https://www.typescriptlang.org/docs/handbook/module-resolution.html) for further information.
 
 ```json
-//tsconfig.json
+// tsconfig.json
 {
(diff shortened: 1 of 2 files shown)


src/i18n/en/docs/module_resolution.md
```diff
@@ -0,0 +1,74 @@
...
+  },
+  "alias": {
+    "react": "preact-compat",
+    "react-dom": "preact-compat",
+    "local-module": "./custom/modules"
+  }
+}
+```
+
+Avoid using `~` in your aliases as it is already used by parcel as mentioned earlier.
```
DeMoorJasper: Users should probably avoid using any special characters all together.

`@` will try to resolve npm orgs
`~` will try to resolve tilde paths

and more characters might be added in the future either from npm, parcel or some third party extension/plugin

src/i18n/en/docs/module_resolution.md
```diff
@@ -0,0 +1,74 @@
+# :notebook_with_decorative_cover: Module Resolution
+
+Parcel (v1.7.0 and above) supports multiple module resolution strategies out of the box so you don't have to deal with endless relative paths `../../`.
+
+## Absolute Paths
+
+`/foo` will resolve `foo` relative to the **project root** (top-level `package.json`)
```
DeMoorJasper: This isn't correct, `project-root` isn't related to `package.json` as `package.json` is an optional file and therefore makes it impossible for us to use it as a project root.

Project root is the directory of the entrypoint or the shared root in case of multiple entrypoints, you can have a look at how it works here: https://github.com/parcel-bundler/parcel/blob/master/src/utils/getRootDir.js

src/i18n/en/docs/module_resolution.md
```diff
@@ -0,0 +1,74 @@
...
+  "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {
+      "~*": ["./src/*"]
+    },
+```
+
+### Monorepo Resolution
+
+TODO: answer this - https://github.com/parcel-bundler/parcel/pull/850#issuecomment-372105317
```
DeMoorJasper: As far as I've tried tilde paths don't work inside aliased packages and monorepos, I might be wrong though.

My advice would be to avoid tilde paths altogether and use relative paths as much as possible and preferably as explicit as possible (use extensions and such). And if you really need a root path, use `/`

jthegedus:
@DeMoorJasper Thanks for the feedback, I've made updates which I think provide sufficient fixes :smiley:

src/i18n/en/docs/module_resolution.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+### Monorepo Resolution
+
+These are the advised usages with monorepos at this time:
+
+Advised usage:
+
+- use relative paths.
+- be as explicit as possible (use file extensions).
```
DeMoorJasper: This wasn't scoped at monorepo's this should be advised across all code as not using extensions causes Parcel to guess the extension possibly linking the wrong file.

Everything else looks good :)

jthegedus: Ah righto. Updated :smiley:

Issue #216, opened 2018-07-30
📃 Add documentation about .env support


pyk:
This [merged feature](https://github.com/parcel-bundler/parcel/pull/258) is very useful, maybe we can add a simple documentation on the "Advanced" section about this feature.

DeMoorJasper:
Feel free to try this and open a PR.

PR #230, opened 2018-08-24
document env file support

jthegedus:
This is my suggestion for documenting `.env` support https://github.com/parcel-bundler/parcel/pull/258.

The original suggestion https://github.com/parcel-bundler/website/issues/216 was for this to be listed under the
>advanced section

but most of the advanced section is about authoring plugins/packagers and using the direct api. I thought keeping it alongside the usage explanation, namely post Code Splitting, a better fit.

I may raise an issue in the future to discuss reorganising the documentation as I've found it difficult for a number of reasons in the past.

### DIFF
--- a/src/i18n/en/docs/env.md
+++ b/src/i18n/en/docs/env.md
@@ -1,0 +1,19 @@
+# 🌳 Environment Variables
+
+Parcel uses [dotenv](https://github.com/motdotla/dotenv) to support loading environment variables from `.env` files.
+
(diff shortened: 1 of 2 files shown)


DeMoorJasper:
I agree that the docs should eventually be reorganized.

Thanks for all your contributions to the docs lately :)

daggerok:
Hi, @keemor 
use https://en.parceljs.org/env.html

PR #235, opened 2018-08-29
Simplified "getting started" code blocks

halo:
The first impression is important. 

I propose these code steps that look like 4 steps be visually reduced to 2.

I did a visual check on the build of every locale to make sure there is no broken Markdown.

---

screen shot 2018-08-29 at 17 24 15

---

### EDIT:
In the PR code it's `# npm` not `# NPM` as in the screenshot.

### DIFF
--- a/src/i18n/en/docs/getting_started.md
+++ b/src/i18n/en/docs/getting_started.md
@@ -4,23 +4,21 @@ Parcel is a web application bundler, differentiated by its developer experience.
 
 First install Parcel using Yarn or npm:
 
-Yarn:
 ```bash
+# Yarn
 yarn global add parcel-bundler
-```
 
-npm:
-```bash
+# npm
 npm install -g parcel-bundler
 ```
 
 Create a package.json file in your project directory using:
 
 ```bash
+# Yarn
 yarn init -y
-```
-or 
-```bash
+
+# npm
 npm init -y
 ```
 
--- a/src/i18n/es/docs/getting_started.md
+++ b/src/i18n/es/docs/getting_started.md
@@ -4,23 +4,21 @@ Parcel es un empaquetador de aplicaciones web, que se diferencia por la experien
 
 Primero instala Parcel utilizando Yarn or npm:
 
-Yarn:
 ```bash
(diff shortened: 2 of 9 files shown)


DeMoorJasper:
I like it, but now it isn't really copiable as they should only run one of those commands not both.

Copying those codeblocks would result in installing using both yarn and npm

halo:
@DeMoorJasper thank you for your feedback!

>  they should only run one of those commands not both

That's why ideally it should look like the following. Default to convention but offer an alternative (that isn't really possible with default Markdown so I opted for what came closest).

screen shot 2018-08-30 at 09 56 35

Another alternative in plain Markdown would be this, but it requires translation:

```bash
npm install -g parcel-bundler

# Or if you use Yarn:
yarn global add parcel-bundler
```

---

> Copying those codeblocks would result in installing using both yarn and npm

It looks like Parcel works under the assumption that people already know what `Yarn` and `npm` are. Given that assumption, they won't run both commands but only one. 

If it's not clear to the users what the difference is, or of there is any doubt about it, there should be no mention at all of `Yarn` and Parcel should just recommend `npm` for starters.

DeMoorJasper:
@halo this makes me lean more towards keeping it as is as I really don’t wanna see yarn go away in the docs. I’d rather see npm go away

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jthegedus:
Just my 2c on converting the docs to use a documentation tool.

While [Gatsby](https://www.gatsbyjs.org/) and things like [Docusaurus](https://docusaurus.io/) are cool, I'd like to see the site built using Parcel. The simplicity it has now is nice, except for the lack of simple things like a dev-server. Maybe using `mdx` with the [parcel-plugin](https://github.com/mdx-js/mdx/tree/master/packages/parcel-plugin-mdx) would be the way to get the best of all worlds?

- simplicity
- parcel DX
- simple JS for solving the problem this issue raised
- doggfooding

Just an idea. I'll be toying with it for another project over the next week. Will link back with thoughts/results :+1:

PR #240, opened 2018-09-05
Add: Ukrainian language

EugeneFenko:


now.json
```diff
@@ -9,6 +9,7 @@
     "pl.parceljs.org",
     "pt.parceljs.org",
     "ru.parceljs.org",
-    "zh.parceljs.org"
+    "zh.parceljs.org",
+    "ua.parceljs.org"
```
devongovett: language code should be `uk` right?

EugeneFenko: Yes, language code: uk, country: ua.

src/i18n/en/index.html
```diff
@@ -45,6 +45,7 @@ <h1>
           <option class="pl" value="https://pl.parceljs.org">Polskie</option>
           <option class="pt" value="https://pt.parceljs.org">Português</option>
           <option class="ru" value="https://ru.parceljs.org">Русский</option>
+          <option class="ua" value="https://ua.parceljs.org">Українська</option>
```
DeMoorJasper: The links should probably get updated as well

PR #250, opened 2018-10-04
Add module resolution translation for pt lang

reebr:


reebr:
Environment variables 🌳 was translated in #243, but does not include in side navigation.
I took advantage of these changes to include the link correctly.

PR #258, opened 2018-10-10
Added infos about cli's "--target" flag regarding package bundling

isthatcentered:
Allows for a more concious choice of export.

I use parcel to bundle npm libraries, discovering "dependencies" are not bundled in "node" export mode was of great use :)

### DIFF
--- a/src/i18n/en/docs/cli.md
+++ b/src/i18n/en/docs/cli.md
@@ -91,7 +91,7 @@ Available in: `serve`, `watch`, `build`
 parcel build entry.js --target node
 ```
 
-⚠️ Target `node` and `electron` will not bundle package.json's `dependencies` (but will include `devDependencies`).
+⚠️ Target `node` and `electron` will not bundle package.json's `dependencies` (but will include `devDependencies`). This behavior can be overriden using --bundle-node-modules flag (see below)
 
 Possible targets: `node`, `browser`, `electron`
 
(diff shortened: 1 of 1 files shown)


src/i18n/en/docs/cli.md
```diff
@@ -91,6 +91,8 @@ Available in: `serve`, `watch`, `build`
 parcel build entry.js --target node
 ```
 
+⚠️ Target `node` and `electron` will not bundle package.json's `dependencies` (but will include `devDependencies`)
```
DeMoorJasper: This can be overwritten using `--bundle-node-modules`. Please add this flag as well and mention it in this warning

isthatcentered: Sure, happy to do so. 

The last two commits address your request :)

src/i18n/en/docs/cli.md
```diff
@@ -91,8 +91,22 @@ Available in: `serve`, `watch`, `build`
 parcel build entry.js --target node
 ```
 
+⚠️ Target `node` and `electron` will not bundle package.json's `dependencies` (but will include `devDependencies`). This behavior can be overriden using --bundle-node-modules flag (see below)
```
DeMoorJasper: Could you use a link for the reference to `#force-node-modules-bundling`

isthatcentered: Oh yeah, duh xD Done

PR #266, opened 2018-10-16
Added info about running Parcel without installing it globally

RapidOwl:
Hi! I was trying to use Parcel with Netlify's build agent and had to do it as described in my change to the Production doc. I wondered if this information might be useful for other people.

I'm sorry I'm not able to localise the text.

### DIFF
--- a/src/i18n/en/docs/production.md
+++ b/src/i18n/en/docs/production.md
@@ -66,4 +66,4 @@ To make working with your project easier, you can add the script to package.json
 }
 ```
 
-You can then build your site with `npm run build`.
+You can then build your site with `npm run build` or `yarn build`.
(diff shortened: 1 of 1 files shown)


RapidOwl:
Thanks! I'll update it this evening.

RapidOwl:
Is that updated script going to work? Installing Parcel as --save-dev won't create a shortcut to the parcel command.

DeMoorJasper:
@RapidOwl npm automatically adds `node_modules/bin` into the execution path, so it knows all commands in node_modules

RapidOwl:
@DeMoorJasper This doesn't work on Windows. Do you mind if I add a bit more for that use case?

DeMoorJasper:
@RapidOwl Alright, forgot windows is always a special case for a minute

PR #272, opened 2018-10-21
docs: add parcel to your project section

jepser:
Add *parcel to your project*, which explains to a add the library specifically to a project so it can run in a CI, instead of installing it globally (which I think is not needed for this type of library).


DeMoorJasper:
This already exists here: https://parceljs.org/production.html#if-you-can't-install-parcel-globally

You might move that to getting started and improve upon it if you want

src/i18n/en/docs/production.md
```diff
@@ -36,34 +36,8 @@ In an effort to optimize production build performance, Parcel will try to determ
...
-To make working with your project easier, you can add the script to package.json.
-
-```json
-"scripts": {
-    "build": "parcel build entry.js",
-}
-```
-
-You can then build your site with `npm run build` or `yarn build`.
+The instructions can be [found here](/getting_started.html).
```
DeMoorJasper: getting_started.html#adding-parcel-to-your-project

jepser: Upps! You're right!

PR #281, opened 2018-10-25
Use Prettier

imbsky:


imbsky:
@DeMoorJasper If you OK with these changes, I will push formatted files.

package.json
```diff
@@ -6,17 +6,31 @@
...
-  },
-  "devDependencies": {
-    "markdown-styles": "^3.1.10",
-    "now": "^11.2.4"
-  },
   "scripts": {
     "build": "./build.sh",
     "clean": "rm -rf dist",
     "start": "node src/server.js",
+    "fmt": "npm-run-all fmt:* -s",
```
DeMoorJasper: This seems a bit overkill. Might be good to just use the same script as we run in parcel's code repo, just one format script

imbsky: Fixed!

PR #285, opened 2018-10-27
Document Flow support

jthegedus:
* Flow section under JavaScript asset type
* Flow section in Module Resolution to demonstrate support of Absolute path mappings in `.flowconfig`

### DIFF
--- a/src/i18n/en/docs/javascript.md
+++ b/src/i18n/en/docs/javascript.md
@@ -1,4 +1,4 @@
-# Javascript
+# JavaScript
 
 _Supported extensions: `js`, `jsx`, `es6`, `jsm`, `mjs`_
 
@@ -53,7 +53,7 @@ const buffer = fs.readFileSync(__dirname + '/test.png')
 
 Install presets and plugins in your app:
 
-```bash
+```shell
 yarn add @babel/preset-react
 ```
 
@@ -65,7 +65,7 @@ Then, create a `.babelrc`:
 }
 ```
 
-## Default babel transforms
+## Default Babel transforms
 
 Parcel transpiles your code with `@babel/preset-env` by default, this is to transpile every module both internal (local requires) and external (node_modules) to match the defined target.
 
@@ -74,3 +74,9 @@ For the `browser` target it utilises [browserslist](https://github.com/browsersl
 The browserlist target defaults to: `> 0.25%` (Meaning, support every browser that has 0.25% or more of the total amount of active web users)
 
 For the `node` target, Parcel uses the `engines.node` defined in `package.json`, this default to _node 8_.
+
+# Flow
+
+[Flow](https://flow.org/) is a popular static type checker for JavaScript. Using Flow with Parcel is as simple as placing `// @flow` as the first line of your `js` files.
+
+Parcel will automatically install the required Babel config to strip the Flow types from the compiled output, so there's nothing you have to worry about except [editor integrations](https://flow.org/en/docs/editors/) or supporting [Absolute Path module resolution](module_resolution.html#flow-/-absolute-or-~-tilde-resolution) with `.flowconfig`.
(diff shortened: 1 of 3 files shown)


src/i18n/en/docs/module_resolution.md
```diff
@@ -66,6 +66,41 @@ and re-export the named export within the aliased file:
 module.exports = require('electron').ipcRenderer
 ```
 
+### Flow / Absolute or ~ Tilde Resolution
```
DeMoorJasper: `Flow with Absolute or Tilde (~) Resolution `?
`Flow / ` might be confusing as it might indicate an explanation about flow inside parcel

DeMoorJasper: Parcel removes flow from bundles automatically so this should become a seperate part in the docs and we don't want confusion between those 2 parts

jthegedus:
```suggestion
### Flow with Absolute or (~) Tilde Resolution
```

I'm confused as this section has the same implications as the following on TS and module resolution with the TS server. This is the same but for the Flow IDE server. Where else would you place this?

src/i18n/en/docs/javascript.md
```diff
@@ -74,3 +74,9 @@ For the `browser` target it utilises [browserslist](https://github.com/browsersl
 The browserlist target defaults to: `> 0.25%` (Meaning, support every browser that has 0.25% or more of the total amount of active web users)
 
 For the `node` target, Parcel uses the `engines.node` defined in `package.json`, this default to _node 8_.
+
+# Flow
+
+[Flow](https://flow.org/) is a popular static type checker for JavaScript. Using Flow with Parcel is as simple as placing `// @flow` as the first line of your `js` files.
+
+Parcel will automatically install the required Babel config to strip the Flow types from the compiled output, so there's nothing you have to worry about except [editor integrations](https://flow.org/en/docs/editors/) or supporting [Absolute Path module resolution](module_resolution.html#flow-/-absolute-or-~-tilde-resolution) with `.flowconfig`.
```
jthegedus:
```suggestion
Parcel will automatically install the required Babel config to strip the Flow types from the compiled output, so there's nothing you have to worry about except [editor integrations](https://flow.org/en/docs/editors/) or supporting [Absolute Path module resolution](module_resolution.html#flow-with-absolute-or-tilde-resolution) with `.flowconfig`.
```

src/i18n/en/docs/module_resolution.md
DeMoorJasper: The typescript one says `ts ~ resolution` not `ts / ~ resolution`

jthegedus: Sorry, I'm not trying to be difficult, I'm just confused. Are you saying the "Flow" section in the `javascript.md` file should be moved into the Babel section?

DeMoorJasper: This https://en.parceljs.org/javascript.html#babel

But this is only for the strip-flow-types default parcel uses the resolution thing you contribute in this PR should remain at the place you put it

jthegedus: Right. I wanted to discuss this topic further in the issue I raised about reorganising the docs.

Most compilation features for JS will be implemented with Babel. If we were to put all those features under the Babel heading it means users need to sift through that whole section just to find that flow types are stripped out. Compared to going to the Flow section to see everything about Flow, how to use, what is done by default and how that is achieved.

I just feel like that's a better experience for the users. It also is more searchable too.

Just my 2c though, happy to revert and propose this semantic change in a different PR.

DeMoorJasper: Yeah a flow section would be great, as long as it's not a single line mentioning Flow types are stripped.
So yeah if you wanna create a page with how to configure it and everything that would be amazing. It's also currently one of the most searched topics in the parcel docs so it would probably help out a bunch of devs :)

jthegedus: I'll open a further discussion point at a later date further covering those ideas.

For now though, what I was trying to describe was that
* Parcel has support for Flow
* It's supported by Babel via strip-types
* You only need `// @flow` to enable it
* If you want to use Absolute Path Module Resolution as well, Flow will complain so and here's the common code to support that

As far as I can tell, this is all Parcel users would need to know about setting up Flow with Parcel. IMO it's a good thing if the explanation for how a feature is implemented is a single sentence. It's a sign Parcel is doing its job correctly. I expect there are going to be other sections as terse in the future.

PR #288, opened 2018-10-30
add examples from examples repo to recipes page

jthegedus:
This would add inline code examples to the recipes page based off of the [parcel-bundler/examples](https://github.com/parcel-bundler/examples) repo.

why do this?:
* encourage use of these examples from the community
* make the examples more discoverable & direct example contributors too
* reduce noise in other repos by clearly showing examples commonly requested
* offer minimal examples for people to easily fork for their reproduction repos 

ISSUES!!

* currently there are bugs in Codesandbox resolving Parcel's Absolute and Tilde module resolution strategies - https://github.com/CompuIves/codesandbox-client/issues/1249

Possible solutions to land inline code examples
1. remove use of absolute/tilde module resolution in these examples
2. add additional examples with specifically showcasing module resolution with a warning about the online playgrounds being broken

Personally, I think solution 2 will best work for now until the aforementioned issue is resolved.

### DIFF
--- a/src/i18n/en/docs/recipes.md
+++ b/src/i18n/en/docs/recipes.md
@@ -1,152 +1,78 @@
 # 🍰 Recipes
 
-## React
+Parcel example projects using various tools & frameworks.
 
-First we need to install the dependencies for React.
+Source can be found at [parcel-bundler/example](https://github.com/parcel-bundler/examples), contributions welcome!
 
-[Blog Post](http://blog.jakoblind.no/react-parcel/)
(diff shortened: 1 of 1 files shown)


jthegedus:
I have implemented the 2. fix for this in https://github.com/parcel-bundler/examples/pull/35. After merging that this should be good to go :smile:

DeMoorJasper:
@jthegedus Codesandbox is running a very old version of Parcel not sure if this is a good idea.

PR #291, opened 2018-11-01
Add less doc

Taym95:


src/i18n/en/docs/less.md
```diff
@@ -1,3 +1,17 @@
 # LESS
 
 _Supported extensions: `less`_
+
+It's not even necessary to install less 😱, parcel will automatically install less if it detects a less file.
```
DeMoorJasper: Probably also good to mention what parcel uses internally to transpile less and how to configure it

PR #320, opened 2018-12-10
fix hidden banner for small screen

abdullahtariq1171:
# Before 
screenshot 2018-12-10 at 6 02 39 pm

# After
screenshot 2018-12-10 at 6 02 54 pm


abdullahtariq1171:
Proposal: How about we make banner `sticky`?

PR #332, opened 2018-12-21
Removed expired T-Shirt/Hoodie Banner

mdchristians:
**Just removed the banner at the top of the homepage about the tshirt or hoodie sale because it expired.**

This PR simply removed the banner all together.  The other option is to just remove the text about the tshirt like in the following image:

screen shot 2018-12-20 at 8 48 22 pm

### DIFF
--- a/src/i18n/en/index.html
+++ b/src/i18n/en/index.html
@@ -23,12 +23,6 @@
     </script>
   </head>
   <body>
-    <div class="banner">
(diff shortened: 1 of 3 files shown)


devongovett:
Thanks. I think it will return shortly though since there have been over 50 requests already...

DeMoorJasper:
Gonna close this for now as they can always come back

PR #335, opened 2018-12-23
Scroll activeLink into view

ZhangYiJiang:
This ensures the current page's link is always visible, which is useful when browsing the documentation and reading each of the page. Otherwise the reader has to keep scrolling down the page when they're reading through the later sections. Personally I would prefer if the scroll state is preserved from page to page, but that takes a lot more code. 

This would have been a one line change, but it seems the templates in other languages were not Prettier formatted, so there's a lot more changes than I intended, unfortunately, and it all got smushed into my existed changes automatically by Husky.

### DIFF
--- a/src/i18n/en/layout/page.html
+++ b/src/i18n/en/layout/page.html
@@ -159,6 +159,7 @@
         var activeLink = document.querySelector('a[href="' + pathname + '"]')
 
         activeLink.classList.add('selected')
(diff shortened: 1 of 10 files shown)


Taym95:
Great work! there is a small issue the title is a little bit hidden.

screenshot 2019-02-21 at 16 35 36

PR #338, opened 2018-12-28
Remember current page on language change

### LINKED ISSUE #301: Switching language redirects to home page
If you change your language you're redirected to a homepage, instead of landing on the same page with a different language.

mischnic:
Closes #301

The change in `src/i18n/en/index.html` and `src/i18n/en/layout/page.html` was applied to all languages, they just weren't formatted previously.


mischnic:
An issue I just noticed: if the selected language doesn't have a corresponding page, an error is printed instead of the landing site.

PR #343, opened 2019-01-04
Update typeScript.md doc for React+JSX

Offirmo:
Additional doc for TS + React + JSX, since it doesn't work out of the box.

### DIFF
--- a/src/i18n/en/docs/typeScript.md
+++ b/src/i18n/en/docs/typeScript.md
@@ -23,3 +23,44 @@ console.log(message)
 // message.ts
 export default 'Hello, world'
 ```
+
+## When using React
+To use Typescript + React + JSX, you need to:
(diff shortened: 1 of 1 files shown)


src/i18n/en/docs/typeScript.md
```diff
@@ -23,3 +23,44 @@ console.log(message)
 // message.ts
 export default 'Hello, world'
 ```
+
+## When using React
+To use Typescript + React + JSX, you need to:
+
+1. use the `.tsx` extension
+2. properly require React
+3. use a tsconfig with a [special option](https://www.typescriptlang.org/docs/handbook/jsx.html) `"jsx": "react"`
```
DeMoorJasper: This shouldn't be necessary afaik

PR #357, opened 2019-01-19
rename all references in docs from 'parcel-bundler' to 'parcel'

amilajack:
Upon importing `parcel-bundler` from node (using parcel's api), `eslint-plugin-import` will fail to resolve the package. This PR fixes that issue


amilajack:
It fails because `eslint-plugin-import` can't resolve it. That may be an issue on the plugin's end tho

DeMoorJasper:
@amilajack sounds like it's either not a dependency or a plugin issue

PR #398, opened 2019-03-04
added link to createapp.dev/parcel

jakoblind:


jakoblind:
@DeMoorJasper Hi, I added a link to a tool that I think is very useful for new Parcel users. Happy to discuss where in the docs it should belong, maybe I put it too high up? Could add a section at the bottom with "external links to tools" or something?

DeMoorJasper:
@jakoblind hey, the tool looks pretty cool. However I don't really see the point of this as Parcel requires nothing except for maybe a package.json.

Also I tried out the tool for a basic React + less app and it adds a lot more dependencies than necessary in particular the loaders are not required by Parcel as these are webpack specific things. Also the babel dependencies are also not necessary except for maybe `@babel/core` to ensure we detect the correct version.

This is actually all you need in this case:
```json
{
  "devDependencies": {
    "parcel-bundler": "^1.12.3",
    "@babel/core": "^7.4.0",
    "less": "^3.9.0"
  }
}
```

jakoblind:
Parcel is awesome because it's zero-config just as you say and many experienced people don't need this tool. However, I know many new devs need this type of resources even though it's simple to setup. Some people are more visual in how they learn. 

Also, this tool is useful for people who haven't made up their minds about whether to use webpack or parcel. They can see that parcel is more simple :) 

About the dependencies: You discovered a bug. There shouldn't be the loader dependencies. Will fix this bug today. Thank you for letting me know

DeMoorJasper:
@jakoblind I actually discovered another bug, at the bottom the cli command also includes webpack for parcel projects

jakoblind:
I just need to publish new version to NPM. Will do it now @DeMoorJasper ! thanks!

PR #400, opened 2019-03-06
update json.md

gandhiShepard:
just a simple example of how to require a json file into our js

### DIFF
--- a/src/i18n/en/docs/json.md
+++ b/src/i18n/en/docs/json.md
@@ -1,4 +1,32 @@
 # JSON
 
 _Supported extensions: `json`, `json5`_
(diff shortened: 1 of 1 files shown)


src/i18n/en/docs/json.md
```diff
@@ -1,3 +1,4 @@
 # JSON
 
 _Supported extensions: `json`, `json5`_
+You can import json files into your js like this `const data = require("filename.json")`.
```
DeMoorJasper: Same as yaml, any import syntax works

gandhiShepard: updates made

PR #401, opened 2019-03-06
update yaml.md

gandhiShepard:
example of how to require a yaml file into your js

### DIFF
--- a/src/i18n/en/docs/yaml.md
+++ b/src/i18n/en/docs/yaml.md
@@ -1,3 +1,19 @@
 # YAML
 
 _Supported extensions: `yaml`, `yml`_
+You can import yaml files into your js like this `const data = require("filename.yaml");`.
+
+```yaml
+# data.yaml
+
+hello:
+  - world
+  - computer
+```
+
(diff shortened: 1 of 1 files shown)


src/i18n/en/docs/yaml.md
```diff
@@ -1,3 +1,19 @@
 # YAML
 
 _Supported extensions: `yaml`, `yml`_
+You can import yaml files into your js like this `const data = require("filename.yaml");`.
```
DeMoorJasper: This is partially correct, you could also use es6 import syntax

`import data from 'filename.yaml'`

gandhiShepard: Interesting. I haven''t been able to get es6 import syntax to work for me. I was under the assumption that it was in working because there was no `export` coming out of the .yaml or even the json file for that matter. I've only been able to `require` json and yaml files into my projects. Am I doing some thing wring here? Please advise. Thx!

gandhiShepard: updates made

PR #402, opened 2019-03-06
update pug LOCALS documentation

gandhiShepard:
so sorry. I ran into some false positives using locals here.  I can whip up a test repo to prove this if necessary.

### DIFF
--- a/src/i18n/en/docs/pug.md
+++ b/src/i18n/en/docs/pug.md
@@ -51,24 +51,18 @@ Let's assume this file structure:
 .
 ├── package.json
 └── src
-    ├── data.js
-    ├── index.js
-    └── index.pug
+    ├── index.pug
+    └── pug.config.js
 ```
-We need to export a `locals` object and import it into our index.js file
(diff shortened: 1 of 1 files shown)


gandhiShepard:
PLEASE look into making this update ASAP because I have the locals setup wrong. Using a `pug.config.js` file is the only way to set this up. And this IS the only resource on line that shows how to wire this up. 

My mistake initially was that I tried with a `pug.config.js` file first, and it worked. I then moved the `pug.config.js` file out of the Pug directory and dropped it at the root. I then tested to see if the locals would still work if I just use a data.js file to export the locals object. It still worked. BUT, that's because Parcel was still reading from the pug.config.js file at the project root. 

So, the only way to pass locals is from a pug.config.js, `.pugrc` or a `.pugrc.js` file as it's set up here https://github.com/buckle2000/parcel/blob/297fd835e4722e2d5df632fdee3755d7f23f5ee0/src/assets/PugAsset.js. 

I'm soooo sorry, but I got a false positive and didn't know. Locals only work from pug.config.js, `.pugrc` or a `.pugrc.js` and that file can be located in the directory of the package.json file OR in the directory containing the `index.pug` file.

gandhiShepard:
Cool. Soo, sorry about that previous error. Stay awesome!

Issue #406, opened 2019-03-12
Link to all query string in README.md


kitkatayama:
Currently, there is only URL for ko(Korean) resource under "Changing the displayed language" in README.md.

```
### Changing the displayed language

Change the displayed language using a query string:

E.g.: access the Korean version using the following url: http://localhost:5000/?locale=ko
```

For convenience to developer or translators, it is better to change as follows imo.

```
### Changing the displayed language

Change the displayed language using a query string:

* English: http://localhost:5000/?locale=en 
* Espanol: http://localhost:5000/?locale=es  
* Francais: http://localhost:5000/?locale=fr 
* and so on (I'll add complete list when send pull request)
```

It will be displayed as follows:

--- 

### Changing the displayed language

Change the displayed language using a query string:

* English: http://localhost:5000/?locale=en 
* Espanol: http://localhost:5000/?locale=es  
* Francais: http://localhost:5000/?locale=fr 
* and so on (I'll add complete list when send pull request)

---

what do you think?
Regards,

kitkatayama:
this issue resolved by #407 , thank you for your attention

PR #407, opened 2019-03-13
#406 added links to README.md

kitkatayama:
Add all locals links to README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,17 @@ The website is deployed using [Zeit Now](https://zeit.co/now).
 
 ### Changing the displayed language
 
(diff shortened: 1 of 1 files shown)


mischnic:
Why is this needed? There is a language chooser at the top:

Bildschirmfoto 2019-03-13 um 17 44 21

DeMoorJasper:
@mischnic on local dev that doesn't work

mischnic:
> on local dev that doesn't work

🙈 How could I miss that this changes the readme and not the actual page

kitkatayama:
thank follow-up @DeMoorJasper :) 🙈

PR #415, opened 2019-03-25
Update language on approach to transpiling node_modules

outofambit:
In light of https://github.com/parcel-bundler/parcel/issues/1655, I've updated this language to be more accurate to the current state of parcel.


DeMoorJasper:
This isn't 100% accurate but it is more accurate than the current state I guess

mrcoles:
Are you saying the way it handles dependencies is to do default `@babel/preset-env` on node_modules/ files unless the package.json in that dependency has a "browserslist" or "engines.browsers" key specified in their package.json? I just spent an hour or two debugging why `"browserslist": "Chrome >= 41"` seemed to not be working the [Swiper package](https://github.com/nolimits4web/swiper/).

PR #417, opened 2019-03-28
Update scss.md

kondinskis:
Remove note for issue as it is resolved.

### DIFF
--- a/src/i18n/en/docs/scss.md
+++ b/src/i18n/en/docs/scss.md
@@ -18,4 +18,4 @@ Dependencies in the SCSS files can be used with the `@import` statements.
 
 If you don't have `sass` module installed before running Parcel, Parcel will install it automatically for you.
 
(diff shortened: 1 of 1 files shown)


src/i18n/en/docs/scss.md
```diff
@@ -18,4 +18,4 @@ Dependencies in the SCSS files can be used with the `@import` statements.
 
 If you don't have `sass` module installed before running Parcel, Parcel will install it automatically for you.
 
-**Notes:** You can also use `node-sass` module for SCSS compilation. Using `node-sass` module will give you faster compilation. However, [an issue](https://github.com/parcel-bundler/parcel/issues/1836) has been reported using `node-sass` module with Parcel.
```
DeMoorJasper: It hasn't been resolved has it?

We just switched to `sass` internally for Parcel which does not cause multithreading issues.
Not sure if the note is even correct though. `node-sass` is not always faster afaik as `sass` is more stable with multiple processes it might actually be faster

PR #428, opened 2019-04-23
Update vue example to support HMR

blanchg:
As it was before the HMR would throw an error when editing the .vue file


blanchg:
This change may be invalid as it appears the doco was changed on purpose to the new method.  Not sure what I am missing though.

PR #458, opened 2019-06-14
Clarify process.env caveat in documentation

awhitty:
Noticed this was a bit unclear and sunk some time into debugging why `console.log(process.env)` wasn't showing any values until I found the [referenced issue](https://github.com/parcel-bundler/parcel/issues/2299). Hopefully this will save someone time in the future :)


gravitypersists:
Chiming that this piece of info would have been helpful. My linter (airbnb config) automatically wants me to destructure: `const X = process.env.X` to `const { X } = process.env` and that doesn't work, sending me on a bit of a goose chase.

PR #460, opened 2019-06-18
doc(plugins): add link to npmjs.com

simon04:


mischnic:
That won't give a very exhaustive list of plugins (if the author forgot to add `parcel-plugin` to the keywords - for example I didn't do that).

DeMoorJasper:
@mischnic is right... It should filter on name not keywords we have a required prefix so any npm package starting with parcel-plugin is probably a parcel plugin

Issue #485, opened 2019-08-13
Missed translation for page leads to unnecessary redirect


Drapegnik:
Hi! If page has no translation it redirects to homepage, instead of showing english version:

![](https://i.gyazo.com/d0fecf1f43649a9749dbbc7bda9d663a.gif)

Also it's really annoying that selected language doesn't save to cookie, so I have to keep switching from `russian` to `english` everytime

PR #486, opened 2019-08-13
Fallback english redirect

Drapegnik:
Fix for #485


mischnic:
>  Pretty sure the server.js is never used as Zeit uses their own server implementation to host our static docs.

`server.js` is used, how would Zeit otherwise know that `it.parceljs.org/getting_started.html` should be rewritten to `dist/it/getting_started.html`

PR #500, opened 2019-09-20
Add htmlnano documentation

### LINKED ISSUE #499: How to stop HTML comments being removed on build
When building my application all HTML comments are removed.

I've tried searching the docs and couldn't find anywhere stating that comments are removed at build.

For example, this comment would be removed:

```html
<!-- BODY -->
```
----------
### Some context
I'm trying to server-side render a ReactJS application which I would replace this comment with. I realise I could use a string like 'replace-me' but it seems a hack around and may get messy down the road.

### The question
How would I keep HTML comments after building my application?

mischnic:
Closes #499


mischnic:
@DeMoorJasper The now deploy error went away now, I guess it was a bug on their side.

PR #501, opened 2019-09-22
Update packagers.md

theghostyced:
A simple change that kept on bugging me 🙂


mischnic:
This is actually on purpose, because "Asset" is a class and "`Asset`s" refers to "multiple object of class `Asset`".

PR #515, opened 2019-10-20
Update recipes.md with setup for Svelte

alex-parra:
Added setup steps for Svelte in-line with the existing content for React, Preact, Vue, etc.
Added to bottom but suggest it's pulled up to show next to the other frameworks.

### DIFF
--- a/src/i18n/en/docs/recipes.md
+++ b/src/i18n/en/docs/recipes.md
@@ -223,3 +223,52 @@ Then, in your `index.html` file, add a reference to your JavaScript entry point.
 ```
 
 Done!
+
+## Svelte
+
+First we need to install the dependencies for Svelte.
+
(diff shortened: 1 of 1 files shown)


alex-parra:
Hi @DeMoorJasper 
I can surely add the full details. Just didn't because most of the others don't so thought that was the intended scope. React, Preact and Vue don't include any content for the `index.html` nor the `main.js`. 

What are your thoughts on this?
Is it a goal to expand the content of React, Preact, etc...?
If I add the full setup for Svelte it will become far longer than the others.
And if so, can I expand the others to my best knowledge?

DeMoorJasper:
@alex-parra ow sorry, it used to have a full description with code and everything a while back.

But I checked and it's the same for the other ones I'll merge everything as is

PR #537, opened 2019-11-13
Move Parcel to ZEIT Now v2

coetry:
This PR contains a custom GitHub Action that invokes the build script to generate the static sites, generate a custom `now.json` file with the appropriate locale alias (`es.parceljs.org ...`), and then invokes a `deploy.js` script that leverages `now-client` to make the actual deployment. 

All that's needed for the action to work is the appropriate ZEIT Now token, which can be added to this GitHub repo as a `NOW_TOKEN` environment variable under "Settings". 

The benefits of this approach is that each locale site is served directly from the edge across all of our regions. No express server is required to serve these assets, which significantly improves performance as well as reliability. Visitors from around the world will get native speeds and native language. Furthermore, the `server.js` file is not needed since each site will be available as a subdomain from the root domain and routing is handled by our proxy, not express.


src/server.js
```diff
@@ -1,41 +0,0 @@
-const express = require('express')
```
devongovett: We'll probably need some replacement for this for local development. The `start` script in package.json still references this.

coetry: @devongovett ahh i forgot about the `start` script. I'll replace that with a `dev` script for local dev that spins up a dev server.

devongovett:
Question: I assume we'll lose out on the nice Github comments on PRs with a preview link?

coetry:
@devongovett with this particular workflow, yes you would be losing the GitHub comments and preview link unfortunately. The links would be available in the logs of the GitHub Actions after they run successfully. It's just a trade-off that needs to be made to support this particular method of internationalization, but at the end of the day, I incline towards things that yield the best performance and availability :smile:

package.json
```diff
@@ -8,8 +8,8 @@
   "license": "MIT",
   "scripts": {
     "build": "./build.sh",
+    "dev": "if [ -d \"dist\" ]; then cd dist; serve; else ./build.sh; cd dist; serve; fi;",
```
DeMoorJasper: Could this change be reflected in the README?

mischnic:
Looks like we will need this sooner rather than later, a deployment just failed: https://github.com/parcel-bundler/website/commit/09e52590db2dde92fd63d6f09ec8f1e37da49ef6#commitcomment-38278723
> The option `--prod` is not supported for Now 1.0 deployments. To manually alias a deployment, use `now alias` instead.

DeMoorJasper:
@mischnic maybe just merge this and improve dev experience ourselves?

mischnic:
👍

mischnic:
Can we merge this? I think Devon needs to set the env variable?

DeMoorJasper:
@mischnic we can merge it. The env variable can be set later

mischnic:
> we can merge it

Can you merge it even though the checks fail?

DeMoorJasper:
@mischnic oh no, we need to disable the github action for this repo...

PR #545, opened 2019-11-25
Update api.md

ericmorand:
It looks like it is possible to configure Parcel to log nothing by passing 0 as the `logLevel` option value.

TypeScript definitions are wrong though. They onle declare 3, 2 and 1 as possible values. They need to be fixed.


ericmorand:
> This is true, we do not maintain the typescript definitions ourselves though.

I will provide a PR at DefinitelyTyped. Looks like the definitions are a bit outdated - and not quite comfortable to use.

PR #550, opened 2019-12-16
add module/nomodule pattern to getting started

AnthumChris:
Would it be helpful to recommend the module/nomodule pattern to target modern browser optimizations?  I'm not sure if this is the best practice by providing ES Module bundles.


mischnic:
Parcel 2 isn't released yet and that website only covers Parcel 1 (so far).

AnthumChris:
> Parcel 2 isn't released yet and the website only covers Parcel 1 (so far).

Ok thanks, I'll close until then.

PR #557, opened 2020-01-18
Update docs to mention required TypeScript types

madyanalj:
When setting up HMR, the TypeScript compiler would complain as it isn't aware of Parcel's module API. To fix this, types has been added ([see this issue](https://github.com/parcel-bundler/parcel/issues/1063)).

This PR updates the docs so that people are aware of this and don't end up getting the same error.


madyanalj:
No worries! Thanks for the quick merge :)

PR #579, opened 2020-04-03
Add a recipe to transpile async functions

laruiss:
Add a minimal babel configuration example to transpile async functions.

### DIFF
--- a/src/i18n/fr/docs/recipes.md
+++ b/src/i18n/fr/docs/recipes.md
@@ -272,3 +272,24 @@ Ensuite, dans votre fichier `index.html`, ajoutez une référence à votre point
 ```
 
 C'est fait !
+
+## Transpiler les fonctions async
+
+Parcel utilise babel et fait un maximum sans configuration. Néanmoins, pour transpiler les fonctions [async](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Instructions/async_function), vous pouvez rajouter un fichier `.babelrc` avec ce contenu :
+
+```json
+{
+  "presets": [
+    "@babel/preset-env"
+  ],
+  "plugins": [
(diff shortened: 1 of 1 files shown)


src/i18n/fr/docs/recipes.md
```diff
@@ -272,3 +272,24 @@ Ensuite, dans votre fichier `index.html`, ajoutez une référence à votre point
...
+Parcel utilise babel et fait un maximum sans configuration. Néanmoins, pour transpiler les fonctions [async](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Instructions/async_function), vous pouvez rajouter un fichier `.babelrc` avec ce contenu :
+
+```json
+{
+  "presets": [
+    "@babel/preset-env"
+  ],
+  "plugins": [
+    ["@babel/plugin-transform-runtime", {
+      "regenerator": true
```
benavern, line 287: What are those lines for ? I think it is not required for `@babel/preset-env` to work well.

By the way I think it is related to this error https://github.com/parcel-bundler/parcel/issues/2936 ? I've faced it too and can't find a way to fix it... is it the solution ?

laruiss: Hi @benavern, sorry for the delay.
It is to be able to transpile properly async functions.

Without it, I get this error in the console :

```
ReferenceError: regeneratorRuntime is not defined
```

And obviously the code does not work.

benavern: Ah ok. I have struggled with this error too but not with async functions. If I remember well this is when using a bundle (generated by parcel) as a dependancy in a project that uses parcel too.
And I don't have any async function in it.
If you wonna help on it: https://github.com/benavern/z-components

DeMoorJasper:

PR #581, opened 2020-04-06
Add dark-mode

benavern:
Hi, I am a big fan of "dark-mode" in softwares and websites.

I happen to be using a lot the Parcel bundler these days and have experienced some eyes aches when switching from my editor to the documentation website.

So, here is my implementation of the website with "dark-mode" (based on the media query `prefers-color-scheme`)

Hope it will please you! :)


DeMoorJasper:

PR #584, opened 2020-04-12
destructuring assignment only with one variable

malekbenz:
we can only use destructuring assignment only with one variable.


DeMoorJasper:

PR #593, opened 2020-05-14
Add "PARCEL_WORKERS" note for building inside CI

andreid-asd:
While building a medium sized project in CircleCI, the build step would almost every time fail with "Out of memory" due to the automatic CPU scaling (they use 36 CPU machines).

I imagine this is not a specific issue to CircleCI  and would make sense to reference `PARCEL_WORKERS` also in this section.


DeMoorJasper:

PR #605, opened 2020-06-19
Dead links

mischnic:


DeMoorJasper:
Maybe we should add a precommit hook for linting?

PR #606, opened 2020-06-19
Styling tweaks and basic mobile support

DeMoorJasper:
This PR adds a basic breakpoint for tablets (which should also somewhat work for mobile)
It also tweaks styling slightly on larger devices and some super minor bugfixes in styling

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "nullthrows": "^1.1.1"
   },
   "lint-staged": {
-    "*.{md,scss,json}": "prettier --write"
+    "*.{md,scss,js,json}": "prettier --write"
   },
   "husky": {
     "hooks": {
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "watch": "eleventy --watch",
     "serve": "eleventy --serve --quiet",
     "debug": "DEBUG=* eleventy",
-    "lint": "prettier --write 'src/**/*.{json,md,scss}'",
+    "lint": "prettier --write 'src/**/*.{js,json,md,scss}'",
     "generate-api-docs": "node generate-api-docs/ ../ ./api"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


mischnic:
Shorter lines are easier to read on wide screens (This is just an 16:9 screen, not even an ultrawide)

website-git-styling devongovett vercel app_getting-started_migration_

v2 parceljs org_getting-started_migration_

DeMoorJasper:
@mischnic I'll just move the navbar like Rollup and make the content shrink, looks a lot better than the webpack docs

mischnic:
(I'm also not quite sure if the navbar should be fixed like Rollup's or just on the same scroll context as we have it at the moment)

DeMoorJasper:
@mischnic I think it's good enough now, I didn't do any of the `overflow` and scrolling things like Rollup but I'm pretty sure that's a bit overkill for now.

Also added a way to hide navigation blocks but not entirely sure if that was actually a good idea.

mischnic:
- on desktop, the spacing between the header and `h1` is gone
- I think the collapsible navbar could be great on mobile (because we have more pages than the v1 docs). Could we enable the collapsing only on mobile (and have all sections be minimized by default)? On desktop it doesn't really work since the navbar width changes when pages with long titles are minimized.

mischnic:
And now the different lines on the page have different colors 😄 
`border-bottom: 1px solid var(--border-color);`

DeMoorJasper:
@mischnic they are... but luckily you don't really see it in non-dark mode so it should be fine. Having no border would make the toggle jump the page down a pixel

mischnic:
`border-color: transparent` 😉

DeMoorJasper:
@mischnic I know how to fix that I just don't think we need to, looks fine to me. Unless you really hate it, in which case I'll change it.

mischnic:
...
Bildschirmfoto 2020-06-19 um 14 53 07

DeMoorJasper:
@mischnic changed it anyway. I think I originally misunderstood the comment, if you meant the color from the sidebar border and top border is different. I totally agree it looks better if they're the same and I fixed it to be that way now...

mischnic:
>  if you meant the color from the sidebar border and top border is different.

Yes 😄

PR #616, opened 2020-06-24
Fix module resolution order

znerol:
The following sentence suggests that the resolution order is different when scope hoisting is is disabled:

> Without scope hoisting however, main is preferred to module for better performance:

However, the following list is identical to the previous one (if scope hoisting is enabled). I did not check the actual implementation, so changing the order in the list to match what the sentence is saying might not be the correct fix.


mischnic:
It actually is different:

https://github.com/parcel-bundler/parcel/blob/cc5b4862699a9020c0931268e1c6e5b0437a6179/packages/resolvers/default/src/DefaultResolver.js#L18-L26

Could you change it to source, browser, main, module, index.js for scope hoisting?

znerol:
> Could you change it to source, browser, main, module, index.js for scope hoisting?

That's what the PR is doing. I just checked the source vs. the PR and I believe the proposed changes to the docs are inline with what the implementation does.

PR #621, opened 2020-07-03
Fix arguments order in ReactDom.render calls in recipes/react.

aureooms:


mischnic:
I mix them up *every* time...

PR #622, opened 2020-07-03
Add blog section

mischnic:
- [x] Add a "Docs / Blogs" chooser in the header

- Migrate two Medium blog posts for Parcel 2


DeMoorJasper:
I don't think this should be part of the sidebar navigation, we should probably split this up into a marketing page, blog articles and the docs part

mischnic:
I agree, we'll "just" have to create a marketing/landing page.

PR #626, opened 2020-07-11
Document global alias and externals

101arrowz:
Documents the behavior of global aliases and externals (which were parcel-bundler/parcel#4853 and parcel-bundler/parcel#4072 respectively).


mischnic:
`includeNodeModules` is documented here already: https://v2.parceljs.org/configuration/package-json/#includenodemodules, but it does make sense to put this into module resolution as well. Could you link the two places together?

> but they will instead be `require`d at runtime.

should be "imported". ES6 doesn't have `require`...

And the target `main` is a bad example because the default for `main` is to not include node_modules anyway. Could you just use `app` instead?

101arrowz:
@mischnic Check the latest deployment, I've fixed those issues.

PR #634, opened 2020-07-21
Document Vue

101arrowz:
Documents the functionality added by parcel-bundler/parcel#4892.

Unfortunately, Prism.js does not support Vue syntax highlighting, so I had to hack it in with HTML highlighting.


DeMoorJasper (changes requested):
Could you create the files for the fix link comments, and just write in them TODO: Write this page or something like we did with javascript and pretty much all other languages.

Also I'm not sure if all of this should be a recipe, vue is actually also a language of itself in the parcel 1 docs we made a language/asset type page for it instead of a recipe. You can still add a recipe as a quick getting started with vue doc though

101arrowz:
I was thinking of Vue as a parallel to React rather than a full-blown language. I'll move it to the languages section, but at that point you could call JSX a language of its own and move React to the languages category as well. Do you want me to document SFC syntax in `languages` and other stuff in `recipes`?

src/recipes/react.md
```diff
@@ -8,7 +8,7 @@ eleventyNavigation:
 
 Compared to Webpack, Parcel's paradigm is to use your HTML file as the entry point, not merely the main script:
 
-{% sample "index.html" %}
+{% sample %}
```
mischnic:
```suggestion
{% sample "parcel index.html" %}
```
This was supposed to show that Parcel should be pointed to the HTML entry, as opposed to Webpack for example. I guess I changed to template for `sample` at some point and missed this.

mischnic:
At least for me, a "language" means "a specific filetype/extension" for Parcel. Recipes are examples that don't really contain new information by themselves but make it easier to get started.

With that definition, Vue should be a language.


- Is there some official documentation regarding blocks that we can link to?
- Is `.vuerc` an official config file or is it only for blocks (for the loader)? If it's the former then it should probably have it's own section as well (and a link to the Vue docs).

src/languages/vue.md
```diff
@@ -0,0 +1,218 @@
+---
+layout: layout.njk
+eleventyNavigation:
+  key: languages-vue
+  title: <img src="/assets/lang-icons/vue.svg" alt=""/> Vue
+  order: 4
+---
+
+Note that Parcel 2 _does not support Vue 2_ if you plan to use SFCs; you must use [Vue 3 beta](https://github.com/vuejs/vue-next) or later.
```
mischnic:
```suggestion
Note that Parcel does not support using SFCs with Vue 2, you must use [Vue 3 beta](https://github.com/vuejs/vue-next) or later.
```

src/languages/vue.md
```diff
@@ -0,0 +1,218 @@
...
+  };
+</script>
+```
+
+{% endsamplefile %}
+{% endsample %}
+
+## HMR
+
+Parcel 2 uses the same HMR injection as the official [vue-loader](https://github.com/vuejs/vue-loader) for Webpack, so you'll have a fast, reactive development experience.
```
mischnic:
```suggestion
Parcel uses the same HMR injection as the official [vue-loader](https://github.com/vuejs/vue-loader) for Webpack, so you'll have a fast, reactive development experience.
```

101arrowz:
Custom blocks are not part of the Vue core but are instead an opinionated feature added for vue-loader. The way I implemented custom blocks is _completely different_ (vue-loader loads your script at build time as opposed to runtime), but most things that can be done with vue-loader can be done in my implementation. That's why I don't link to the actual docs for custom blocks.

Also, no, the official `vue-loader` does not use a `.vuerc`, it gets config from `webpack.config.js`.

DeMoorJasper (approved):

mischnic:
> Custom blocks are not part of the Vue 

- Would it make sense then to briefly describe what Custom Blocks are (for someone who only knows Vue core)?

- Are any of the other settings of https://cli.vuejs.org/config/#vue-config-js apply useful/used by the Parcel transformer?

101arrowz:
The vast majority of Vue+Parcel users will have used `vue-loader` at some point (since that's what Vue CLI uses) and will know what custom blocks are. Moreover, the example I provided is pretty self-explanatory.

None of the options in `vue.config.js` really apply to Parcel - Vue CLI is meant to be a very simplified skin over webpack, so nearly all of the options (e.g. `pwa.workboxPluginMode`) are not relevant and should be specified elsewhere.

mischnic:
> vast majority will know what custom blocks are.

> The way I implemented custom blocks is completely different

I think they should be described in more detail

101arrowz:
Implementation is different, concepts are the same. It's just code injected at runtime to modify the component instance. The example I provided is very self-explanatory. If you want, I'll make another PR though.

mischnic:
Alright, let's leave it as is until someone opens an issue asking for more details

PR #645, opened 2020-08-18
Update cli.md with a note about glob patterns

Kangaroux:
Bit of a gotcha with regards to glob patterns that I think is worth calling out.

If you have a file tree and you run `parcel build ./**/*` without any extension it won't find the files correctly. `parcel build './**/*'` works as expected.

### DIFF
--- a/src/features/cli.md
+++ b/src/features/cli.md
@@ -26,6 +26,16 @@ parcel a.html b.html
 parcel ./**/*.html
 ```
 
+NOTE: If a glob pattern ends in an asterisk `*`, be sure to wrap the glob in single quotes, otherwise you may experience issues caused by your shell.
+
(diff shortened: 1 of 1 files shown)


mischnic:
1. I would put this just before the `parcel [serve] <entries>` heading
2. Please do something like this:

```
{% warning %}

...The text you added with "NOTE"...

{% endwarning %}
```

3. This is recommended with any glob (not just those ending in an asterisk) because it ensures that files which were added while Parcel is running are also picked up without needing to restart (this should also be mentioned).

Kangaroux:
@mischnic Updated

mischnic (approved):

DeMoorJasper (approved):

PR #647, opened 2020-08-23
[T-558] validator docs

DeMoorJasper:
This PR adds more complete docs to the validator page


src/plugin-system/validator.md
```diff
@@ -7,35 +7,83 @@ eleventyNavigation:
...
 ```
 
-Some validators (such as `@parcel/validator-typescript`) may wish to maintain a project-wide cache for efficiency. For these cases, it is appropriate to use a different interface where parcel hands _all_ changed files to the validator at the same time:
+## How to write a stateful validator
+
+Some validators (such as `@parcel/validator-typescript`) may wish to maintain a project-wide state/cache for efficiency. For these cases, it is appropriate to use a different interface where parcel hands _all_ changed files to the validator at the same time.
+
+For this validator type Parcel also ensures to always invoke this validator on the same thread (so that your cache state is accessible).
+
+This type of validator is way slower than the stateless validator type as it runs everything on a single thread, instead of multiple threads. Only use this if you really have no other choice, this is usually the case for validators that need to have access to the entire project, like Typing validators.
```
mischnic:
```suggestion
This type of validator is typically slower than the stateless validator type as it runs everything on a single thread, instead of multiple threads. Only use this if you really have no other choice, as is usually the case for validators that need to have access to the entire project, like Typing validators.
```

src/plugin-system/validator.md
```diff
@@ -7,35 +7,83 @@ eleventyNavigation:
 summary: "A plugin type: Analyze assets and emit warnings and errors"
 ---
 
-Validators receive an asset, and can throw errors if that asset is invalid
-in some way, e.g. type errors or linting errors.
+A validator is a plugin type that is used to validate assets, it gets each asset as an input and based on that asset you can throw errors or log warnings with diagnostics to ensure the input is as it should be. This can be linting issues, typing issues or any other validation issue.
```
mischnic:
```suggestion
A validator is a plugin type that is used to validate assets, it is called on each asset and can throw errors or log warnings with diagnostics to validate the input. These can be linting issues, typing issues or similar.
```

"A validator validates" is still a bit strange though

src/plugin-system/validator.md
```diff
@@ -7,35 +7,83 @@ eleventyNavigation:
...
+
+Validators run after a build has fully completed, this is to ensure we don't impact performance and focus on more important compilation errors first.
+
+When Parcel is being ran in watch mode, so `parcel watch` or `parcel serve` we still serve/save updated bundles even when a validator throws an error. We log the error and nothing else really happens. However when running `parcel build` we do exit the process with a failure status code to ensure you don't deploy any code that does not meet the criteria set out by your validators. This ensures developers can stay productive and not have too worry about every small typing or linting issue while trying to solve a problem.
```
mischnic:
```suggestion
When Parcel runs in watch mode (`parcel watch` or `parcel serve`), we still serve/save updated bundles even if a validator throws an error (in this case the error is merely logged).

But when running `parcel build`, Parcel exists with a failure status code to ensure you don't deploy any code that does not meet the criteria set by your validators. This ensures developers can stay productive and don't have to worry about every small typing or linting issue while trying to solve a problem.
```

src/plugin-system/validator.md
```diff
@@ -7,35 +7,83 @@ eleventyNavigation:
...
+
+## How to write a stateless validator plugin
+
+If a validator shouldn't store any state about the application you can utilise the standard validator plugin format, this gives you one asset at a time allowing us to spread the work across threads, which improves performance.
+
+These validators receive an asset, and can throw errors or log a warning if that asset is invalid in some way, e.g. type errors or linting errors.
+
+To do this properly you should be using diagnostics, these are described in the logger docs under [`diagnostics`](/plugin-system/logger/#diagnostics).
+
+Below you can find a simple example on throwing an Error
```
mischnic:
```suggestion
Below you can find a simple example of throwing an Error.
```

mischnic:
For the stateful validator, it doesn't say that this cache is just a top-level variable (and the example doesn't have one either).

And the logger call in `validateAll`  doesn't use `assets` but `asset`

mischnic:
In case you haven't seen my comment: https://github.com/parcel-bundler/website/pull/647#issuecomment-683411442

DeMoorJasper:
@mischnic thanks for the heads up, I've seen the comment but forgot to update this PR, will probably look into this tonight.

mischnic:
No worries!

PR #713, opened 2020-10-01
Update html.md

BigPackie:
Added documentation for usage of  PostHTML  **posthtml.config.js**   config file with examples ( _posthtml-modules_, _posthtml-shorten_ ) and notes about  caveats.


DeMoorJasper (approved):

BigPackie:
Now all that remains is to wait for somebody to merge it?

DeMoorJasper:
@BigPackie yes, but unfortunately we can't merge if Vercel deployment fails... Not sure why though as I can't see the logs.

BigPackie:
Is there somebody you can contact to solve this?

DeMoorJasper:
@BigPackie fixed it, your usage of sample was incorrect apparently

BigPackie:
Thx a lot.  Can you share me a link about correct usage of those sample tags, so I don't make the same mistake next time?

BigPackie:
One more thing, I realized the string `{{headerTitle}}` does not show up in the website within my example. That's a problem. How is it possible to make it visible?

mischnic:
> so I don't make the same mistake next time?

I always copy them from another page.
Also: the whole point of these tags was to have the complete, self-contained example in a single "code sample" box and not distributed across multiple paragraphs.

PR #729, opened 2020-10-07
Update html.md

BigPackie:
Trying to fix an issue with not displaying '{{headerTitle}}' on the page. Improved example.


BigPackie:
Still problematic, now it displays '\\{\\{headerTitle\\}\\}' isntead of escaping them.

BigPackie:
The best I could do to display '{{ }}' correctly was to put a space in between them, like this '{ { } }'. However, this is still not a complete solution. Does anybody have an idea?

DeMoorJasper (approved):

PR #732, opened 2020-10-09
Improve website design

EduardoLopes:
This PR is still a working in progress. The goal is to improve the website design overall, fixing small but important things, mainly on the dark theme.

- [x] Improve color in the dark theme
- [x] Improve font colors (it's too contrasty on the dark theme)
- [x] Use sans-serif in the whole website (This makes font usage more consistent)
- [x] Make the theme switch icon clear on light theme (show something more close to a sun)
- [x] On the sidebar, make the clickable itens different from not clickable itens
- [x] Improve the style of the links in the top bar (docs and blog links)


mischnic:
🎉

EduardoLopes:
@mischnic there's any reason the contents inside .main-content uses a serif font?

mischnic:
The v1 docs used a serif font as well: https://parceljs.org/javascript.html 🤷‍♂️ 

I think that the font stack you chose is rather hard to read (too bold/too small), I'd prefer something like  this: `font-family: "Helvetica Neue",Helvetica,Arial,sans-serif;`, looks like you applied the title font to the content as well?

---

>  Make the theme switch icon clear on light theme (show something more close to a sun)

Just so we're on the same page: when I designed that button, my basic idea was modelling a solar eclipse (which also explains the css transition when switching):
- in light mode: the moon is about to cover the sun (and incidentally, this also looks like a crescent moon): 
https://en.wikipedia.org/wiki/File:Partial_solar_eclipse_Oct_23_2014_Minneapolis_5-36pm_Ruen1.png
- in dark mode: the moon is covering the sun, leaving a small ring of the sun visible: 
https://en.wikipedia.org/wiki/File:Annular_Eclipse._Taken_from_Middlegate,_Nevada_on_May_20,_2012.jpg

Maybe we can add some kind of halo that resembles a halo

https://commons.wikimedia.org/wiki/File:Solar_eclipse_1999_4_NR.jpg

and a stylistic sun at the same time (like the crescent moon):

Bildschirmfoto 2020-10-15 um 16 13 36

Or do you have other ideas?

EduardoLopes:
hmm, didn't notice v1 used a serif font. To have a better readability when using a serif-font, we should increase the font-size a little bit, this would be better for read on dark background too. But i think a sans-serif font makes more sense for Parcel and it's better for readability over all, in the current font size at least.

I've set the the title font to the content just to see how it would look, and end up committing, as it was looking good to me. But we can set the font-family you mentioned, if you agree to keep the sans-serif font.

<hr/>

About the theme toggle, i noticed it was a eclipse when i was modifying yesterday. As a concept this is really cool, but it's a little confusing for the final user. I have my system set to dark mode as default and when i first entered the website, i didn't notice it was a button, i thought it was the open collective icon (as it's aside of the github icon). Only when i changed my system theme to light, i noticed it was a moon, then i noticed that was the theme toggle button.

I pushed the changes i made, check it out and tell me what do you think. In my opinion, now it looks more like a sun, and it's easier to identify as a theme toggle.

mischnic:
> it's a little confusing for the final user

Yeah, I can always try to rework that at some later time. At least the functionality is obvious now.

> But we can set the font-family you mentioned, if you agree to keep the sans-serif font.

👍 

---

Some things I noticed:
- the smooth color change for the siderbar/content border line and the docs/blog button isn't covered by the various `transition: background ....`
- What do you think about not applying any opacity to the sidebar headers? In light mode, this makes barely any difference but in dark mode, they look a bit too gray-ish. I think just having a different font and no hover should be enough to mark them as non-interactive.

mischnic:
We probably want to use the system font by default: https://nathanfriend.io/2018/04/27/fantastic-fast-fonts-with-system-ui.html

```css
font-family: -apple-system, system-ui, BlinkMacSystemFont, 
             'Segoe UI', Roboto, 'Helvetica Neue', 
             Ubuntu, Arial, sans-serif;
```

And there's a merge conflict now 😬

EduardoLopes:
I've made the changes you point out and fixed the conflict. I'll take a look into the smooth color change for the siderbar/content later.

mischnic:
The sidebar should definitely still use the title font as previously. Not sure about the headings in the content, probably also the old title font?

EduardoLopes:
Since now the font is diferente on each system, i was thinking about font matching when i set everything to use the the same font. But i just push a revert for that. It looks good, on windows, at least!

mischnic:
👍 
macOS:

Bildschirmfoto 2020-10-16 um 21 32 41

mischnic:
We can also merge this PR without the hamburger menu and the docs/nav switcher change (but fixing the smooth color-scheme transition for the sidebar border and the subheading text color would be great), there already another merge conflict...

EduardoLopes:
I was doing a technical test for a selective process for a job and I didn't have time to finish this PR. Sorry.

But i have some free time now. I'll make sure to finish it this week. And do the menu stuff in another PR.

mischnic:
This can wait, no need to hurry 😄

EduardoLopes:
Sorry it took me so much time to finish this PR. As i said before, i was doing a selective process for a job and end up getting the job, so this week was my first week working as a front-end developer. 🎉🎉

I removed the menu task, as you said, and fixed the side bar transition and stuff. If you feel like merging this (if approved), you can do already. ✨

mischnic:
Congratulations!

One last thing: before, the docs/blog switcher made it apparent which section you're currently viewing.

Bildschirmfoto 2020-11-13 um 23 15 38

After your latest changes, both the active and the inactive link have the same style, change you change it so that the active link is highlighted as in the first screenshot for both dark & light mode (so that they are always blue-ish).

Bildschirmfoto 2020-11-13 um 23 16 55

<br>

(And please add `..., border-right-color 0.2s ease-in-out;` here:
https://github.com/parcel-bundler/website/blob/5d9c46205c3dc5fda6f2abe7640a9bb66b486982/src/doc-styles.scss#L21-L23
that should make the border between navbar and content also change gradually when switching themes)

src/doc-styles.scss
```diff
@@ -35,21 +35,26 @@ html {
   --sidebar-color: #333;
   --code-mark-background-color: #ff0;
   --border-color: #eee;
-  --hover-nav-background: #f8f8f8;
-  --selected-nav-background: #eee;
+  --hover-nav-background: #e5e5e5;
+  --selected-nav-background: #d8d8d8;
+  --font-color-1: rgba(0, 0, 0, 0.97);
+  --font-color-2: rgba(0, 0, 0, 0.7);
+  --font-color-3: rgba(0, 0, 0, 0.48);
```
mischnic: I think these are now unused? (both here and below in `html.dark`)

EduardoLopes:
why this is set to scroll?
https://github.com/parcel-bundler/website/blob/5d9c46205c3dc5fda6f2abe7640a9bb66b486982/src/doc-styles.scss#L521-L523

here's how it looks on windows:
Screenshot 2020-11-14 130651

mischnic:
If I remove that, I can't scroll the main content itself (if it contains more text)

mischnic:
I think it was just a typo, this works for me:

```diff
diff --git src/doc-styles.scss src/doc-styles.scss
index 40c6399..b51d595 100644
--- src/doc-styles.scss
+++ src/doc-styles.scss
@@ -420,7 +420,7 @@ body {
   font-size: 1.1rem;
   white-space: nowrap;
   @media (min-width: $breakpoint-tablet) {
-    overflow-y: scroll;
+    overflow-y: auto;
   }

   .doc-nav-group-hidden {
@@ -518,8 +518,9 @@ body {
   min-width: 0;
   line-height: 1.5em;
   box-sizing: border-box;
+
   @media (min-width: $breakpoint-tablet) {
-    overflow-x: scroll;
+    overflow-y: auto;
   }

   & > div {
```

mischnic:
And I forgot that this should be `transition-colors`, so that the border color transition is actually applied 🙈 

https://github.com/EduardoLopes/website/blob/5571441a78546f1b4268ea7631f45adb6f0ee92d/src/doc-styles.scss#L427

EduardoLopes:
Done ✨

mischnic (approved):
🎉   Thanks!

@DeMoorJasper Looks good to you?

PR #736, opened 2020-10-14
Make nav bar scrollable and set max width for content

alvyjudy:
I made some small change to the stylesheet to improve the UI of the website:) hope it helps. 

There doesn't seem to be any guidelines on how the stylesheet should be written but do let me know if I'm breaking any conventions.

### DIFF
--- a/src/doc-styles.scss
+++ b/src/doc-styles.scss
@@ -64,6 +64,8 @@ body {
   display: flex;
   flex-direction: column;
 
+  overflow: hidden;
+
   @media print {
     font-size: 12pt;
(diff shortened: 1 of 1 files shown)


mischnic:
That's a great idea, but there are some problems (for example https://website-2rbqdp3bs.vercel.app/getting-started/library/ has no min-width and both scrollbars overflow so that the "Edit this page" link is hidden).

I've taken a slightly different approach in this PR: https://github.com/parcel-bundler/website/pull/740. Thanks again for the inspiration!

PR #743, opened 2020-10-16
Use CSS grid for overall layout

mischnic:
I had hoped this would fix the anchor scrolling problem in Chrome/Safari (click on this link: https://v2.parceljs.org/features/module-resolution/#externals, Chrome has scrolled the HTML element, the top bar is out of view - Firefox works fine though), but it didn't

Maybe we still want this? This changes the nested flexbox situation `(header, (sidebar, content)` to a grid
```
H H H
S C C
```
(_H_eader, _S_idebar, _C_onent)

### DIFF
--- a/src/_includes/layout.njk
+++ b/src/_includes/layout.njk
@@ -63,65 +63,63 @@
         {% include "darkmode.html" %}
       </div>
     </header>
-    <div id="content" class="content-container">
-      <nav class="docs-navigation">
-      {% if tags and tags.includes("Blog") %}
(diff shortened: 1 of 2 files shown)


mischnic:
The scroll problem was fixed by removing `position: absolute` from the visually hidden class: [`74adba0` (#743)](https://github.com/parcel-bundler/website/pull/743/commits/74adba061d988ff17bd2e10fb7357214c75d171b) 🤷

DeMoorJasper (approved):

PR #750, opened 2020-10-22
Fix import in API proxy section

ranisalt:
Since `reporter-dev-server` depends on `http-proxy-middleware ^0.19.1`, the import is incorrect. From `^1.0.0`  onwards, it would be correct, as the API was changed.

### DIFF
--- a/src/features/api-proxy.md
+++ b/src/features/api-proxy.md
@@ -38,10 +38,23 @@ This would cause `http://localhost:1234/api/endpoint` to be proxied to `http://l
 For more complex configurations, a `.proxyrc.js` file allows you to attach any (connect-compatible) middleware, this example has the same behaviour as the `.proxyrc` version above.
 
 {% sample %}
+{% samplefile "package.json" %}
+
+```json
+{
+  "devDependencies": {
+    "http-proxy-middleware": "^1.0.0",
+    "parcel": "next"
+  }
+}
+```
+
+{% endsamplefile %}
+
 {% samplefile ".proxyrc.js" %}
 
 ```js
-const createProxyMiddleware = require("http-proxy-middleware");
+const { createProxyMiddleware } = require("http-proxy-middleware");
 
 module.exports = function (app) {
   app.use(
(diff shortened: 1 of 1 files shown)


DeMoorJasper:
@ranisalt it's probably better to update the dependency in Parcel core instead?

mischnic:
What we should really do is add a `package.json` to the docs that specifies the user's version of that package.
Both to make Yarn PnP happy and to prevent breakage when we update the version in the reporter.

ranisalt:
> What we should really do is add a `package.json` to the docs that specifies the user's version of that package.

You mean as a peer dependency? I imagine several uses will not have this installed. I think it would be better to re-export it from parcel so it will always use the dependent version.

mischnic:
No, add it as a dev dependency in your project (so next to `"parcel": "^2.0.0-...`).

ranisalt:
> No, add it as a dev dependency in your project (so next to `"parcel": "^2.0.0-...`).

But then it would have to be a (versionless?) peer dependency of dev-server since it uses HPM, otherwise module resolution might try to use different versions.

mischnic:
No, no peer dependency.
The config file is executed in the same way as `node .proxyrc.js`. So the dependencies are resolved relative to the config file: if you have `http-proxy-middleware ` as a dependency/devDependency in your project's `package.json`, it will use that (whatever version).
(The dev-server does not take your config and execute it as though it was located at `node_modules/@parcel/reporter-dev-server/.parcelrc.js`)

If you don't specify it at all, you are relying on Yarn/npm's hoisting to give you *some* version. It could also happen that you still import version `0.20.0` in your config although Parcel itself uses `1.0.0`. And there are als package managers (Yarn PnP, pnpm?) that intentionally throw an error if you try to require a dependency that isn't declared.

If you think about a babelrc file: the package names you specify in there are also required from your project, and you should explicitly list the presets/plugins in your dependencies.

ranisalt:
Hmm that makes it clearer to me.

mischnic:
Can you add a package.json file with `http-proxy-middleware` listed as a dev dependency to the second example on that page?  Like here: https://v2.parceljs.org/languages/html/#inline-style-attribute

ranisalt:
@mischnic there you go!

mischnic (approved):
Thanks!

PR #761, opened 2020-11-12
Added documentation that there will always be at least one target, even with `main: false`

### LINKED ISSUE #759: Document that there will always be at least one target, even with `main: false`
This doesn't mean that Parcel will do nothing when running `parcel build src/index.js` but that the default target will be used instead of `main`.
```json
{
  "main": "unrelated.js",
  "targets": {
    "main": false
  }
}
```

https://github.com/parcel-bundler/parcel/issues/5312#issuecomment-721999509

AceTheCreator:
This pull request added documentation that there will always be at least one target, even with `main: false` #759


src/configuration/package-json.md
```diff
@@ -55,6 +55,8 @@ To make Parcel ignore one of these fields, specify `false` in `target.(main|brow
 
 If the `browser` field is an [object](/features/module-resolution/#package.json-browser-field), `package.json#browser[pkgName]` can be used instead of `package.json#browser`.
 
+Note that there will always be at least one target, even with `main: false`(This doesn't mean that Parcel will do nothing when running parcel build src/index.js but that the default target will be used instead of main.)
```
mischnic:
```suggestion
Note that there will always be at least one target, even with `main: false`. (So the default target will be used instead of `main`.)
```

AceTheCreator: @mischnic I have implemented the suggested changes. please review the pull request.

mischnic (approved):

PR #764, opened 2020-11-14
Dark search popover

### LINKED ISSUE #737: Search results popover: should respect dark mode
https://docsearch.algolia.com/docs/styling/

Bildschirmfoto 2020-10-14 um 13 15 18

mischnic:
Closes https://github.com/parcel-bundler/website/issues/737

Bildschirmfoto 2020-11-14 um 21 58 32

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -14,7 +14,12 @@ module.exports = function (eleventyConfig) {
   eleventyConfig.addPlugin(eleventyNavigationPlugin);
   eleventyConfig.setDataDeepMerge(true);
 
-  eleventyConfig.addPlugin(pluginSass, { watch: ["src/*.{scss,sass}"] });
+  eleventyConfig.addPlugin(pluginSass, {
+    watch: ["src/*.{scss,sass}"],
+    cleanCSSOptions: {
+      level: 0,
+    },
+  });
   eleventyConfig.addPassthroughCopy("assets");
   eleventyConfig.setTemplateFormats(["md", "css", "png", "svg"]);
   eleventyConfig.addWatchTarget("./api/");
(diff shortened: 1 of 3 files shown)


.eleventy.js
```diff
@@ -14,7 +14,12 @@ module.exports = function (eleventyConfig) {
   eleventyConfig.addPlugin(eleventyNavigationPlugin);
   eleventyConfig.setDataDeepMerge(true);
 
-  eleventyConfig.addPlugin(pluginSass, { watch: ["src/*.{scss,sass}"] });
+  eleventyConfig.addPlugin(pluginSass, {
+    watch: ["src/*.{scss,sass}"],
+    cleanCSSOptions: {
```
mischnic: that css minified would also remove the first newly added rule 🤷

src/doc-styles.scss
```diff
@@ -804,3 +804,47 @@ body {
     }
   }
 }
+
+html.dark {
+  .algolia-autocomplete {
```
DeMoorJasper: Can't you add `/* clean-css ignore:start */` above this line and `/* clean-css ignore:end */` at the end of this css block to skip removal of these lines?

Not sure if we should completely opt-out from optimising for this little bit of css

DeMoorJasper (changes requested):

src/doc-styles.scss
mischnic: > Not sure if we should completely opt-out from optimising for this little bit of css

I've reordered the rules and cleancss seems to work correctly now.

The difference was surprisingly small:
disabled: 19kb
level 0: 16kb
level 1 (default): 15kb

DeMoorJasper: oh awesome, probably means we don't have a lot of unnecessary css?

DeMoorJasper (approved):

PR #777, opened 2020-12-03
Add WebExt docs

101arrowz:
Only relevant after parcel-bundler/parcel#5304 is merged.


src/recipes/web-extension.md
```diff
@@ -0,0 +1,56 @@
+---
+layout: layout.njk
+eleventyNavigation:
+  key: recipes-webext
+  title: <img src="/assets/lang-icons/webext.svg" alt=""/> Web Extension
```
mischnic: Maybe add a class `dark-invert` to the stylesheet that applies `filter: invert(1)` if dark mode is enabled? Otherwise the black icon isn't visible on the dark background.

src/recipes/web-extension.md
```diff
@@ -0,0 +1,87 @@
...
+    "webext-dev": {
+      "sourceMap": {
+        "inline": true,
+        "inlineSources": true
+      }
+    },
+    "webext-prod": {}
+  },
+  "scripts": {
+    "dev": "parcel src/manifest.json --host localhost --target webext-dev",
```
mischnic: Please rename to `start` to be consistent with the other doc pages

mischnic:
I have no experience with webextensions, so: Is it the case that importing css/`url:` imports fail because:
1. they are not listed in `web_accessible_ressources` and
2. the URL that Parcel outputs by default would have to be wrapped with `chrome.runtime.getURL(...)` so that it's translated into `web-extension://afe2138/....`.

101arrowz:
The two fail for two different reasons, as far as I can tell. URL imports fail because, as you said, the `browser.runtime.getURL` function must be called on the URL import for them to work. While we could make it possible to use `url:` imports, most users are going to end up following an online resource, and almost every resource on web extensions tells you to use `getURL`. The current system is closer to native Web Extension development for that reason.

The CSS fails because there is no HTML file to attach to in content scripts. Honestly, it's pretty much expected that this happens because you're injecting a stylesheet rather than using it directly. Of course, CSS imports still work in options pages, popup pages, etc. (i.e. where HTML is the root asset). We could solve this by detecting a CSS asset in the subtree/having the CSS asset alert the parent that it exists, then having the Web Extension transformer add the CSS asset to the `css` field? Not sure if this is feasible or not, but either way it's not really necessary for most users.

src/recipes/web-extension.md
```diff
@@ -0,0 +1,87 @@
...
+    "start": "parcel src/manifest.json --host localhost --target webext-dev",
+    "build": "parcel build src/manifest.json --target webext-prod"
+  }
+}
+```
+
+{% endsamplefile %}
+{% endsample %}
+
+Running `yarn run dev` or `npm run dev` will start the development server. HMR and source maps will work for background scripts, content scripts, the popup page, and the options page. To add the extension to your browser, research how to load an extension unpacked (for example, in Chrome, [click "Load Unpacked"](https://developer.chrome.com/extensions/getstarted#manifest)).
```
mischnic:
```suggestion
Running `yarn start` or `npm start` will start the development server. HMR and source maps will work for background scripts, content scripts, the popup page, and the options page. To add the extension to your browser, research how to load an extension unpacked (for example, in Chrome, [click "Load Unpacked"](https://developer.chrome.com/extensions/getstarted#manifest)).
```

src/recipes/web-extension.md
```diff
@@ -0,0 +1,87 @@
+---
+layout: layout.njk
+eleventyNavigation:
+  key: recipes-webext
+  title: <img class="dark-invert" src="/assets/lang-icons/webext.svg" alt=""/> Web Extension
+  order: 7
+---
+
+Web Extension development is enabled by `@parcel/config-webext`.
```
mischnic:
```suggestion
Bundling WebExtensions is enabled by using `@parcel/config-webext` like this:
```

"WebExtension" seems to be the official name for this API

101arrowz:
```suggestion
Bundling WebExtensions is enabled by using `@parcel/config-webextension` like this:
```

src/recipes/web-extension.md
```diff
@@ -0,0 +1,87 @@
...
+
+{% sample %}
+{% samplefile ".parcelrc" %}
+
+```json
+{
+  "extends": "@parcel/config-webext"
```
101arrowz:
```suggestion
  "extends": "@parcel/config-webextension"
```

src/recipes/web-extension.md
```diff
@@ -0,0 +1,87 @@
...
+```json
+{
+  "extends": "@parcel/config-webextension"
+}
+```
+
+{% endsamplefile %}
+{% endsample %}
+
+Running Parcel on your [`manifest.json`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json) will automatically validate and bundle your web extension if you pass it as an entry.
```
mischnic:
```suggestion
Running Parcel with your [`manifest.json`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json) as an entry (`parcel build manifest.json`) will automatically validate and bundle your web extension.
```

mischnic (approved):

mischnic (approved):

PR #782, opened 2020-12-15
Document image transformer

DeMoorJasper:
This PR is an attempt at documenting the image transformer

### DIFF
--- a/src/recipes/image.md
+++ b/src/recipes/image.md
@@ -50,7 +50,8 @@ An HTML example:
   <body>
     <picture>
       <source src="url:./image.jpeg?as=webp&width=800" type="image/webp" />
-      <img src="url:./image.jpeg?width=800" alt="test image" />
+      <source src="url:./image.jpeg?width=800" type="image/jpeg" />
+      <img src="url:./image.jpeg?width=200" alt="test image" />
     </picture>
   </body>
 </html>
--- a/src/recipes/image.md
+++ b/src/recipes/image.md
@@ -1,9 +1,9 @@
 ---
 layout: layout.njk
 eleventyNavigation:
-  key: languages-image
+  key: recipes-image
   title: <img src="/assets/lang-icons/image.svg" alt=""/> Image
-  order: 0
+  order: 2
 ---
 
 Image assets can be imported and processed like any other asset in Parcel. As with any other asset you can import this asset from any other asset, so you can import images from html, css, js, ts, ...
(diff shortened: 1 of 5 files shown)


mischnic:
I think this makes more sense as a "recipe" than a language?

DeMoorJasper:
@mischnic I don't think it belongs in either, not sure but maybe languages should be renamed to `asset types` or something?

mischnic:
Yeah. Generally, I like that languages is more high-level than asset types though.

DeMoorJasper:
@mischnic just checked and in Parcel 1 we actually used to call it asset types, I'm totally fine with adding a new section for `images`, `manifests`, ... just not sure how to call it besides asset types

mischnic:
Can we put in recipes for now? We can always reorganize it once we know what it should be

DeMoorJasper:
@mischnic sure moved it for now,

src/recipes/image.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Parcel image transformer
+
+Using the Parcel transformer `@parcel/transformer-image` you can resize, change the format and quality of an image. To do this we added the possibility to define query parameters.
+
+The query parameters you can use are:
+
+- `width`: The width to resize the image to
+- `height`: The height to resize the image to
+- `quality`: The image quality you want, for example `75`
```
mischnic:
```suggestion
- `quality`: The image quality percentage you want, for example `75`
```

src/recipes/image.md
```diff
@@ -0,0 +1,36 @@
...
+- `format`: File format to use, for example: `webp`
+
+Supported image formats: `jpeg/jpg`, `png`, `webp`, `tiff`, `heic/heif` and `raw`
```
mischnic:
```suggestion
Supported image formats: `jpeg` / `jpg`, `png`, `webp`, `tiff`, `heic` / `heif` and `raw`
```

I think this is what you mean?

src/recipes/image.md
```diff
@@ -0,0 +1,36 @@
...
+An example:
+
+{% sample %}
+{% samplefile "index.js" %}
+
+```js
+import imageUrl from "url:./image.jpeg?format=webp&width=250";
+```
+
+{% endsamplefile %}
```
mischnic, line 35: Should we add an example for HTML (`img src=`) as well? Do you need `url:` in that case?

DeMoorJasper: Sure, will try it out, I think url is still required although not sure, probably depends on the config?

DeMoorJasper: `url:` is required in the default config, so used that for now

mischnic (approved):
🎉

PR #783, opened 2020-12-23
Fix SVG htmlnano documentation

### LINKED ISSUE #441: Need to document htmlnano configuration examples
# Problem

When using parcel for builds, my hand-written svg was being completely re-written, stripped and broken. I had to look through the dependency chain to discover what was happening and I falsely accused posthtml by raising an issue there.

I see that a number of other developers have struck this problem and struggled to configure parcel's environment for htmlnano and svgo.

# Solution (that needs to be documented)

To avoid parcel's toolchain from minifying hand-written SVG, add the following `.htmlnanorc.js` file, delete the `.cache` directory and run a re-build.

## .htmlnanorc.js

```js
module.exports = {
  "minifySvg": false
}
```

## For bonus points

htmlnano is flexible and brings many capabilities. I think more examples on how to configure htmlnano - or at least a few suggestions and then link to htmlnano so people can discover all the options - will help people make the most of *parcel*.

brodul:
Hi,

The htmlnano setting key must be `minifySvg` and not `minifySVG` . This PR probably fixes #441 

Ref: 
https://github.com/posthtml/htmlnano#minifysvg
https://github.com/parcel-bundler/parcel/issues/1523#issuecomment-431613312

### DIFF
--- a/src/languages/html.md
+++ b/src/languages/html.md
@@ -273,7 +273,9 @@ The result should be:
 
 If minification is enabled (e.g. `parcel build` without `--no-minify`) all bundles are automatically processed with [htmlnano](https://github.com/posthtml/htmlnano).
 
-It can be configured according to its documentation with a .htmlnanorc (JSON) or .htmlnanorc.js file, for example to retain HTML comments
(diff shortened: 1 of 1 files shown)


mischnic (approved):
Thanks!

PR #790, opened 2021-01-25
plugin-configuration.md - exampel with "..." in extending config

elwin013:
Example of extending configuration with `"..."` in extending config section based on reporters plugins.

There is usage example of `"..."` in next section, but it could be beneficial (from user perspective) to add
similar info in the `Extending configs` section, as it looks like main place to answer the question 
"how to extend some config by adding custom configuration".

### DIFF
--- a/src/configuration/plugin-configuration.md
+++ b/src/configuration/plugin-configuration.md
@@ -79,7 +79,7 @@ A common usecase is extending the default config, for this reason the `extends` 
 {% endsamplefile %}
 {% endsample %}
 
-If default config need to be extended (for example additional reporters needs to be added, but default ones need to stay), add `"..."` to extend list provided by config specified in `extends` field.
+To add additional plugins while not overwriting the but default ones, add `"..."` to use the plugins provided by config specified in the `extends` field. For example to add an additional reporter:
 
 {% sample %}
 {% samplefile ".parcelrc" %}
--- a/src/configuration/plugin-configuration.md
+++ b/src/configuration/plugin-configuration.md
@@ -87,7 +87,7 @@ If default config need to be extended (for example additional reporters needs to
 ```json/1
 {
   "extends": "@parcel/config-default",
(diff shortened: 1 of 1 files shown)


src/configuration/plugin-configuration.md
```diff
@@ -79,6 +79,22 @@ A common usecase is extending the default config, for this reason the `extends`
...
 
+If default config need to be extended (for example additional reporters needs to be added, but default ones need to stay), add `"..."` to extend list provided by config specified in `extends` field.
+
+{% sample %}
+{% samplefile ".parcelrc" %}
+
+```json/1
+{
+  "extends": "@parcel/config-default",
+  "reporters": ["...", "@parcel/reporter-dev-server"]
```
mischnic: The dev server is already enabled by default...
```suggestion
  "reporters": ["...", "parcel-reporter-custom"]
```

src/configuration/plugin-configuration.md
```diff
@@ -79,6 +79,22 @@ A common usecase is extending the default config, for this reason the `extends`
 {% endsamplefile %}
 {% endsample %}
 
+If default config need to be extended (for example additional reporters needs to be added, but default ones need to stay), add `"..."` to extend list provided by config specified in `extends` field.
```
mischnic:
```suggestion
To add additional plugins while not overwriting the but default ones, add `"..."` to use the plugins provided by config specified in the `extends` field. For example to add an additional reporter:
```

src/configuration/plugin-configuration.md
elwin013: :+1: Yep, also `parcel-reporter-custom` seems better.

src/configuration/plugin-configuration.md
elwin013: Oh, that's sounds better, thanks!

mischnic (approved):
Thanks!

PR #806, opened 2021-02-07
Don't suggest using generateScopedName for css modules

mischnic:
I have a feeling that many people copy the first example for enabling PostCSS modules into their project and needlessly change `generateScopedName`.


mischnic:
We need to fix `Error: PostCSS config must have plugins` with `{"modules": true}` first...

DeMoorJasper (approved):

PR #835, opened 2021-03-18
Add in high level definitions for bundling terms

AGawrys:
Noticed this was TODO. 

These are some high level definitions that could be a place holder until we add more. 
Let me know and I can alter or insert more information

**I am not authorized to merge in this Pull Request. So maybe someone else can or I can be granted access? Thanks!**

### DIFF
--- a/src/plugin-system/data-structures.md
+++ b/src/plugin-system/data-structures.md
@@ -19,7 +19,7 @@ This is not intended to be an API documentation but rather introduces the concep
 Any file supported by Parcel, such as JS, CSS, or HTML
 
 ##### Dependency
-When a file depends on or requires other files, those files are dependencies.
+When a file depends on or requires other files, those imports are dependencies.
 
 ##### AssetGraph
 Graph data structure representing all assets and their dependencies on one another
--- a/src/plugin-system/data-structures.md
+++ b/src/plugin-system/data-structures.md
@@ -16,17 +16,17 @@ This is not intended to be an API documentation but rather introduces the concep
 {% endnote %}
 
 ##### Asset 
-Any file supported by Parcel, such as JS, CSS, or HTML
+Any file supported by Parcel, such as JS, CSS, or HTML.
 
 ##### Dependency
 When a file depends on or requires other files, those imports are dependencies.
 
 ##### AssetGraph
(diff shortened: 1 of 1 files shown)


src/plugin-system/data-structures.md
```diff
@@ -15,14 +15,18 @@ This is not intended to be an API documentation but rather introduces the concep
 
 {% endnote %}
 
-##### Asset
+##### Asset 
+Any file supported by Parcel, such as JS, CSS, or HTML
 
 ##### Dependency
+When a file depends on or requires other files, those files are dependencies.
```
mischnic: What we call dependency (and the type `Dependency`) is actually the "import" "itself", not the dependency resolution (which you're describing here).

mischnic:
There's also some overlap between this page and https://v2.parceljs.org/plugin-system/overview/ 🤷

src/plugin-system/data-structures.md
AGawrys: > There's also some overlap between this page and https://v2.parceljs.org/plugin-system/overview/ 🤷

I see... not really sure what the point of the Data structures page is then, I just saw it empty and figured it would be a quick fix to add in some definitions. So either the definitions should be sprinkled into the overview or other pages, or we're okay with duplicates, or just remove the BundleGraph and Asset graph definitions? What do you think?

>dependency
Maybe a better definition then would be "Dependency - Any imports a file contains are dependencies" ?

@mischnic

src/plugin-system/data-structures.md
```diff
@@ -15,14 +15,18 @@ This is not intended to be an API documentation but rather introduces the concep
...
+When a file depends on or requires other files, those imports are dependencies.
 
 ##### AssetGraph
+Graph data structure representing all assets and their dependencies on one another.
 
 ##### Bundle
+Groupings of assets that get placed together in a single file. Bundles will (generally) contain only assets in the same language.
 
 ##### BundleGraph
+Graph data structure graphing all assets, bundles, and their dependencies.
```
mischnic:
```suggestion
Graph data structure containing bundles including their assets and dependencies.
```

mischnic (approved):

PR #837, opened 2021-03-20
Update code-splitting.md

Andrew-Dyachenko:
Possible TYPO at
`({ x } = await import("./b.js"));`
Did you mean:
`let ({ x } = await import("./b.js"));` ?


mischnic:
No, this is actually correct (and valid JS).

PR #844, opened 2021-03-24
Added missing quotes to the package.json examples

gautamkrishnar:
Added missing quotes to the package.json examples, qutoes were missing and it will cause errors to the users who are copy pasting the code from the site.


mischnic (approved):
Thanks

PR #880, opened 2021-06-27
Update preact module resolution example

wei:
preact compact is now under the main preact package as `preact/compat`.


DeMoorJasper (approved):

DeMoorJasper:
we can no longer update v1 docs unfortunately

PR #885, opened 2021-07-08
Update api-proxy.md

shawsourav:
We need to use default export from "http-proxy-middleware" to make this work.


mischnic:
This depends on the version of `http-proxy-middleware` you're using. If you add it as a dependency as recommended in that example (the package.json just before the JS code), it's a named import.

PR #886, opened 2021-07-10
TypeScript support declaration is misleading

michalzubkowicz:
Only transpiling is supported. TypeScript is all about typings not transpiling. People expect that some toolkit elements have same behaviour without knowledge whats under the hood if you ignore typings it works in diffrent way that in the rest of ecosystem.

### DIFF
--- a/src/languages/typescript.md
+++ b/src/languages/typescript.md
@@ -7,7 +7,7 @@ eleventyNavigation:
 summary: Explains the different ways TypeScript can be transpiled
 ---
 
-Typescript works out of the box with Parcel, but you have multiple options:
+TypeScript type checking doesn't work. Only transpiling is supported, but you have multiple options:
 
 ## Babel
 


mischnic:
The current description regarding what's enabled by default is indeed misleading, but there is a section about how to enable type checking on that page.

So it should rather say something like "Transpiling TypeScript works out of the box with Parcel"

michalzubkowicz:
Junior devs don't know the difference and will expect that transpiling
means "everything".
It's really bad that what you try to do here is insistently false
marketing.

More honest declaration should be "TypeScript support doesn't work",
because TypeScript is all about types and You just skipped that feature,
but there is a hack to enable it.

devongovett:
Sorry but that's just wrong. Many tools work exactly like Parcel does. Transpiling means transpiling and nothing more. And you _can_ enable type checking if you want.

PR #902, opened 2021-08-15
Docs updates, part 1

### LINKED ISSUE #421: Clarify "Bundle Resolution"
The section on Bundle Resolution under Code Splitting seems to describe something that is clever and interesting  :
> Bundle resolution
> Parcel infers the location of bundles automatically. This is done in the bundle-url module, and uses the stack trace to determine the path where the initial bundle was loaded.

and also provides advice that seems important:
> This means you don't need to configure where bundles should be loaded from, but also means you must serve the bundles from the same location.

As a beginner Parceler, I have little idea of what is actually being  described or why this is part of the Code Splitting section.  Perhaps some context, examples, use case, and/or motivation material could be added to help clarify what this section is about.  

Thanks!

### LINKED ISSUE #756: Module resolution: extension priority
Document extension priority: https://github.com/parcel-bundler/parcel/blob/26a67e4aa9693a029a5c6edd6eb91b461ef8f6f6/packages/utils/node-resolver-core/src/NodeResolver.js#L110-L114

### LINKED ISSUE #765: Document the image transformer (resizing, different format)
https://github.com/parcel-bundler/parcel/blob/v2/packages/core/integration-tests/test/integration/image/reformat.js

### LINKED ISSUE #843: Describe how to use named pipelines with various other tools
- Typescript: ambient module
- Flow: rewrite & a dummy JS file
- jest: same as Flow?

devongovett:
Part 1 of some docs updates I'm working on.

* Adds getting started guide for libraries
* Updates the getting started guide for apps to simplify it a bit, and make it agnostic to React.
* Combines package.json configuration page, and parts of (no) configuration page into a page about Targets along with a bunch of new info.
* Combines the HMR page and Proxy server page into a page on Development, along with a bunch of new stuff.
* Significantly expands the Module Resolution page (now called Dependency resolution)
* Adds much more detail to the Production page
* Adds a Bundle Inlining page

More to come in subsequent PRs.

Closes #421 Closes #756 Closes #765  Closes #843


src/getting-started/webapp.md
```diff
@@ -1,153 +1,155 @@
...
 {% sample %}
-
 {% samplefile "package.json" %}
 
-```json
+```json/3
 {
   "name": "my-project",
+  "source": "src/index.html",
+  "browserslist": ">= 0.25%",
```
mischnic: I image there will be people who just copy this example into their project and I think `>= 0.25%` isn't a good default nowadays? https://github.com/browserslist/browserslist#best-practices

src/features/dependency-resolution.md
```diff
@@ -0,0 +1,427 @@
...
+.logo {
+  background: url(logo.png?width=400&height=400);
+}
+```
+
+See the [Image transformer](/recipes/image/) docs for more details on images. You can also use query parameters in custom [Transformer](/plugin-system/transformer/) plugins.
+
+{% note %}
+
+**Note**: query parameters are not supported for CommonJS specifiers (created by the `require` function).
```
mischnic:
```suggestion
**Note**: Query parameters are not supported for CommonJS specifiers (created by the `require` function).
```
Maybe?

src/features/dependency-resolution.md
```diff
@@ -0,0 +1,427 @@
...
+      "global": "$"
+    }
+  }
+}
+```
+
+{% endsamplefile %}
+{% endsample %}
+
+## Configuring other tools
```
mischnic, line 391: Should this also cover named pipelines (because now they are also described on this page)? 

https://github.com/parcel-bundler/website/issues/843

src/features/development.md
```diff
@@ -0,0 +1,201 @@
...
+
+#### `.proxyrc.js`
+
+For more complex configurations, a `.proxyrc.js` file allows you to attach any [connect](https://github.com/senchalabs/connect)-compatible middleware. This example has the same behaviour as the `.proxyrc` version above.
+
+{% sample %}
+{% samplefile ".proxyrc.js" %}
+
+```js
+const { createProxyMiddleware } = require("http-proxy-middleware");
```
mischnic, line 138: This should mention that the user should add `http-proxy-middleware` as a dependency in this case. Otherwise they rely on whatever version we used in the server reporter

src/features/production.md
```diff
@@ -1,24 +1,109 @@
...
+
+```javascript
+if (process.env.NODE_ENV !== 'production') {
+  // Only runs in development and will be stripped in production builds.
+}
+```
+
+See the [Node emulation docs](/features/node-emulation/) for more details on environment variable inlining.
+
+### Image optimization
```
mischnic, line 50: This makes it sound as though the image transformer only runs in production builds...

src/features/targets.md
```diff
@@ -0,0 +1,320 @@
...
+“Differential bundling” is the idea of shipping multiple versions of your code for different targets, and allowing the browser to choose the most optimal one to download. When you use a `<script type="module">` element in an HTML file, and some of the browsers specified by the environment do not support ES modules natively, Parcel will automatically generate a `<script nomodule>` fallback as well.
+
+```html
+<script type="module" src="app.js"></script>
+```
+
+```html
+​<script type="module" src="app.c9a6fe.js"></script>
+<script nomodule src="app.f7d631.js"></script>
+```
```
mischnic: Something like "[code block] will be turned into [code block]" would be clearer

mischnic (commented):
- Should the information that `browser`, `main`, `module` are only meant for libraries be more prominent? We constantly get issues about that
- Did intentionally leave this section out?
https://github.com/parcel-bundler/website/blob/1adaba652c1a7dd0891806aa5414558fe951942b/src/configuration/package-json.md#L256-L268

src/getting-started/webapp.md
devongovett: hmm yeah. maybe I'll go for `"> 0.5%, last 2 versions, not dead"`, which is pretty close to the defaults. I don't want to just write "defaults" because that doesn't really show how to configure it.

src/features/production.md
devongovett: hmm yeah... it was meant more as a quick mention since it is an optimization. The actual docs for it are on a different page.

devongovett:
> Should the information that browser, main, module are only meant for libraries be more prominent? We constantly get issues about that

Yeah maybe. What did you have in mind? I was pretty careful to only mention those under the "Library targets" section.

Also, I intentionally got rid of usage of `npm init`/`yarn init` in the getting started guides, which should help here as well hopefully.

devongovett:
> Did intentionally leave this section out?

TBH I found this section a bit confusing, so I want to re-think it a little. Maybe we should have a separate page about monorepos?

src/features/dependency-resolution.md
devongovett: Yeah this section needs more expanding still. e.g. the TS example maps ~ to the project root, but that doesn't actually match what Parcel does... I'm not sure if it's really possible though.

devongovett: hmm or should we add a section like this to the Bundle Inlining page?

devongovett: ok added a part here, and a new section in the inlining page to link to it.

src/features/targets.md
```diff
@@ -0,0 +1,322 @@
...
+
+Library targets do not bundle dependencies from `node_modules` by default. In addition, minification is disabled by default for libraries. These can be overridden using the appropriate option in the `targets` field (see below). Scope hoisting cannot be disabled for library targets.
+
+Library targets automatically output either native ES modules or CommonJS depending on the target.
+
+- **`main`** – by default, outputs CommonJS. If the `.mjs` extension is used, or the `"type": "module"` field is specified, then an ES module is output instead.
+- **`module`** – outputs an ES module.
+- **`browser`** – a browser-specific override of the `main` field. Outputs CommonJS.
+
```
mischnic, line 206: Should this mention the `types` target? That is also "special"

mischnic:
> TBH I found this section a bit confusing, so I want to re-think it a little. Maybe we should have a separate page about monorepos?

Sure, just wanted to make sure this doesn't get lost.

> I was pretty careful to only mention those under the "Library targets" section.

True, I guess this is fine for now.

src/features/production.md
mischnic: That does make sense, maybe mention that this always runs? This page does start with:

> Parcel’s production mode ...

src/getting-started/webapp.md
mischnic: There's another instance of this here: https://website-git-docs-devongovett.vercel.app/features/targets/#package.json%23browserslist

mischnic (approved):

PR #904, opened 2021-08-21
Update and extend docs for JavaScript, CSS, and HTML

devongovett:


src/languages/javascript.md
```diff
@@ -0,0 +1,401 @@
...
+```javascript
+exports.multiply = function(a, b) {
+  return a * b;
+};
+```
+
+{% endsamplefile %}
+{% endsample %}
+
+In addition to `require` and `exports`, Parcel also supports `module.exports`, as well as the `__dirname` and `__filename` module globals, and `process.env` for access to environment variables. See the [Node emulation](/features/node-emulation/) docs for more details.
```
devongovett, line 78: I'm debating whether the node emulation docs should actually be merged into this page since they are JS-specific

DeMoorJasper (approved):

src/languages/javascript.md
```diff
@@ -0,0 +1,401 @@
...
+### JSX
+
+Parcel supports JSX out of the box. JSX is automatically enabled in a `.jsx` or `.tsx` file, or when one of the following libraries is defined as a dependency in your package.json:
+
+- `react`
+- `preact`
+- `nervjs`
+- `hyperapp`
+
+The correct JSX pragma is also automatically inferred based on the library you use. If none of the above libraries are installed, then the pragma defaults to `React.createElement`. Parcel also automatically detects the version of React or Preact that is installed, and enables the [modern JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) if supported.
```
mischnic: > If none of the above libraries are installed, then the pragma defaults to `React.createElement`

But in that case, JSX won't be enabled automatically anyway?

src/languages/javascript.md
```diff
@@ -0,0 +1,401 @@
...
+```
+
+{% endsamplefile %}
+{% endsample %}
+
+### Flow
+
+[Flow](https://flow.org) support is automatically enabled when `flow-bin` is listed as a dependency in your project's root package.json. You must also use a `@flow` directive in the files you wish to be compiled.
+
+Parcel currently uses Babel to strip flow types. If you have a custom Babel config, you will need to add the Flow plugin yourself. See Babel for more details.
```
mischnic: > See Babel for more details.

Should this be a link?

src/languages/css.md
```diff
@@ -0,0 +1,246 @@
...
+body {
+  background: url('images/background.png');
+}
+```
+
+Referenced files should be [relative](/features/dependency-resolution/#relative-specifiers) to the containing CSS file. You can also use [absolute](/features/dependency-resolution/#absolute-specifiers) and [tilde](/features/dependency-resolution/#tilde-specifiers) specifiers. The `data-url:` scheme can also be used to inline a file as a data URL. See [Bundle inlining](/features/bundle-inlining/) for more details.
+
+```css
+.logo {
+  background: url(data-url:./logo.png);
```
mischnic:
```suggestion
  background: url('data-url:./logo.png');
```

devongovett: pretty sure quotes aren't required, but I'll add them to match the other example

src/languages/javascript.md
devongovett: yeah I guess I was getting at when you used a `.jsx` or `.tsx` file but didn't have one of those libraries installed

devongovett: but it's kinda confusing in this paragraph so I'll remove it

PR #911, opened 2021-08-28
removing old version

cekvenich2:


devongovett:
Thanks, but I think we should keep the v1 site installing v1. The v2 site is a work in progress and hopefully we'll cut over to it soon by default.

PR #922, opened 2021-09-02
Expand docs on Parcel API

devongovett:
Tried using Node ESM for the examples, which is made a little nicer by https://github.com/parcel-bundler/parcel/pull/6866.

### DIFF
--- a/src/features/parcel-api.md
+++ b/src/features/parcel-api.md
@@ -11,10 +11,10 @@ The Parcel API can be used to programmatically run builds or watch a project for
 
 ## The `Parcel` constructor
 
-The Parcel API can be used through the `@parcel/core` package. First, you need to install it into your project:
+The Parcel API can be used through the `@parcel/core` package. You'll also need a default configuration, such as `@parcel/config-default`.
 
 ```shell
-yarn add @parcel/core
+yarn add @parcel/core @parcel/config-default
 ```
 
 Next, import this package into your program and call the `Parcel` constructor. It accepts an [`InitialParcelOptions`](/plugin-system/api/#InitialParcelOptions) object, which contains all of the options used by the Parcel CLI and a few more.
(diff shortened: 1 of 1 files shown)


src/features/parcel-api.md
```diff
@@ -1,94 +1,295 @@
...
+
+{% sample %}
+{% samplefile "build.mjs" %}
+
+```javascript
+import {Parcel} from '@parcel/core';
+
+let bundler = new Parcel({
+  entries: 'a.js',
+  defaultConfig: '@parcel/config-default'
```
devongovett, line 35: This works if the build script is within the project. Tried to describe this above that it's resolved relative to the project root. It's a lot simpler to show it this way than doing the whole `require.resolve` dance, especially in ESM. Not sure about it though.

mischnic: > This works if the build script is within the project.

But not in the setup as described, `@parcel/core` doesn't have a dependency on `@parcel/config-default`. It would be fine if that was added to the `yarn add` call

src/features/parcel-api.md
```diff
@@ -1,94 +1,295 @@
...
+      browsers: ['last 1 Chrome version']
+    }
+  }
+});
+```
+
+{% endsamplefile %}
+{% endsample %}
+
+When set to an array, the `targets` option can be used to specify which of the project’s targets to build. For example, to only build the project’s `modern` target:
```
mischnic:
```suggestion
When set to an array, the `targets` option can be used to specify which of the project’s targets to build (as described in `package.json`). For example, to only build the project’s `modern` target:
```

(or something like that)

src/features/parcel-api.md
```diff
@@ -1,94 +1,295 @@
...
+{% samplefile "build.mjs" %}
+
+```javascript
+import {Parcel} from '@parcel/core';
+import {fileURLToPath} from 'url';
+
+let bundler = new Parcel({
+  entries: 'a.js',
+  defaultConfig: '@parcel/config-default',
+  additionalReporters: [
```
mischnic, line 154: Should one set `patchConsole: true` in this case?

devongovett: 🤷🏻‍♂️ I don't think core or any of the default plugins use console.log...

src/features/parcel-api.md
devongovett: Right yeah. Depends if we assume the build script is within the project or separate

PR #925, opened 2021-09-05
More docs for .parcelrc, plugin config loading, and Transformers

devongovett:
* Updates Parcel configuration page to add more details on each plugin type
* Adds a section covering the `loadConfig` method of most plugin types
* Adds more docs and example for Transformers

### DIFF
--- a/src/plugin-system/transformer.md
+++ b/src/plugin-system/transformer.md
@@ -4,38 +4,271 @@ eleventyNavigation:
   key: plugin-system-transformer
   title: Transformer
   order: 3
-summary: "A plugin type: Convert an asset (into another asset)"
 ---
 
-Transformers _transform_ single assets as they are discovered and added to the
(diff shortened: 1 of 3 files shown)


src/plugin-system/transformer.md
```diff
@@ -4,38 +4,271 @@ eleventyNavigation:
...
+      specifierType: 'esm'
+    });
 
-  async generate({ asset, ast, resolve, options }) {
-    // ...
-    return { code, map };
+    return assets;
   }
 });
 ```
```
devongovett, line 274: Should we have a section on symbols? I wasn't sure exactly how to cover this...

mischnic: I think that would make more sense as a separate page (like "Logging" at the moment)? Because most transformers won't use them

mischnic (approved):
😍

PR #930, opened 2021-09-07
Docs for Reporters, diagnostics, and logging

devongovett:
Docs for reporters, plus some updates to the diagnostics and logging docs


src/plugin-system/logging.md
```diff
@@ -2,125 +2,163 @@
...
+Once you're familiar with the [`Diagnostic`](#Diagnostic) format, you can log anything you want, from verbose messages to errors with code frames and hints. This example shows how to log a warning, complete with a code frame, hints, and a documentation URL.
+
+```js
+import {Transformer} from '@parcel/plugin';
+
+export default new Transformer({
+  async transform({asset, logger}) {
     // ...
+
+    logger.warn({
```
mischnic, line 138: Should this mention that the warning will only be displayed once when first transforming but then on cached builds if the transformer didn't run?

PR #931, opened 2021-09-09
Add RSS feed

### LINKED ISSUE #896: Add RSS feed to blog
It would be awesome to add an RSS feed to the Parcel blog so that we can follow along using feed readers.

canac:
Closes #896


mischnic:
The title seems to be missing from the xml file: https://website-git-fork-canac-rss-feed-devongovett.vercel.app/feed.xml (notice the empty `<title/>` tags)

What I see when adding the feed to an RSS reader:

Bildschirmfoto 2021-09-19 um 20 27 54

canac:
@mischnic Good catch, thanks! Fixed in 41c6c54

mischnic:
Can you please resolve the merge conflicts? That should also fix the deployment error

devongovett (approved):

PR #932, opened 2021-09-11
Minor bump in usability

jeff-hykin:
From #6887 
- add npm install command
- add notes for edgecase of `npm init -y` causing the package.json to have a main that conflicts with parcel

### DIFF
--- a/src/features/parcel-api.md
+++ b/src/features/parcel-api.md
@@ -15,6 +15,7 @@ The Parcel API can be used through the `@parcel/core` package. You'll also need 
 
 ```shell
 yarn add @parcel/core @parcel/config-default
+npm install @parcel/core@next @parcel/config-default@next
 ```
 
 Next, import this package into your program and call the `Parcel` constructor. It accepts an [`InitialParcelOptions`](/plugin-system/api/#InitialParcelOptions) object, which contains all of the options used by the Parcel CLI and a few more.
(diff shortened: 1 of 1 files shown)


src/features/parcel-api.md
```diff
@@ -39,6 +40,8 @@ let bundler = new Parcel({
 {% endsamplefile %}
 {% endsample %}
 
+Note: If your package.json contains a "main" field, such as `"main": "index.js"` that is going to be used as the output destination by default. If no main is specified, then Parcel will default to creating a `dist/` folder.
```
devongovett, line 43: I don't think this is the right place for this. It has nothing to do with the Parcel API. The Targets page already discusses these fields.

src/features/parcel-api.md
```diff
@@ -15,6 +15,7 @@ The Parcel API can be used through the `@parcel/core` package. You'll also need
 
 ```shell
 yarn add @parcel/core @parcel/config-default
+npm install @parcel/core@next @parcel/config-default@next
```
devongovett, line 18: I'd rather not do this and instead change the tags to point to latest. Otherwise we will have to update this again when Parcel 2.0.0 is released.

src/features/parcel-api.md
jeff-hykin: I see what you're saying, and I agree the current comment/info belongs on the target page.

But, from my side a problem still exists, because even having used parcel for years I've been struggling to get even the most basic setup working with parcel 2. The error message around "main" doesn't link to the target page.

As a potential compromise, is there a bundler argument for node.js that sets the output folder/file? When people are copying and pasting the example code, having an output argument that overrides the "main" from the package json would side step the problem caused by `npm init -y`.

PR #933, opened 2021-09-12
Fix typo

Vikaash02:


mischnic:
This is intentional: https://stackoverflow.com/questions/27961879/why-is-using-return-as-the-last-statement-in-a-function-considered-bad-style

Vikaash02:
Understandable, thank you.

Vikaash

PR #935, opened 2021-09-19
Update website styles

devongovett:
This updates the styles for the website using Tailwind. To do this, it runs the output of Eleventy through Parcel, which now handles SASS compilation and CSS processing. Parcel also does some optimization for the site which Eleventy didn't do: minifying JS/CSS, and optimizing SVG and images (the PNGs get a lot smaller).

* Uses Tailwind colors everywhere
* Adds hamburger navigation menu on mobile
* Adds table of contents section for current page headings on large screens, which highlights the current section as you scroll
* Improves accessibility


DeMoorJasper (approved):

mischnic:
- I find the darker text color in the navigation bar in the previous version easier to read (instead of grey)
- The moon variant of the dark mode switcher isn't transparent anymore (the crescent is filled)
- There seems to be a delay/flicker sometimes after changing the page in the navigation bar until the navigation section switches back to the previous scroll position (https://github.com/parcel-bundler/website/blob/0eee269/src/_includes/remember-nav-scroll.js)
- This is a bit of an edge case, but the elements in the header intersect on my phone

devongovett:
Ugh looks like svgo broke the moon icon. 😞 

Not sure what to do about the intersecting elements. It's pretty tight up there. Do we start hiding things on mobile?

Which text color were you concerned with? Inside the left nav or the top nav?

mischnic:
😬 

Yeah, the `color="black"` part is somehow stripped

https://github.com/parcel-bundler/website/blob/0eee269b833cfac6322e072968a4af8bd0bada8e/src/_includes/darkmode.html#L5

> Not sure what to do about the intersecting elements. It's pretty tight up there. Do we start hiding things on mobile?

Probably? The alternative would be wrapping, which is what we do currently (though that's not ideal eitehr).

>  Inside the left nav or the top nav?

On the left: `.docs-navigation a`

devongovett:
Yeah I could probably darken it a little. Wanted to have enough headroom for the hover style, but then I added a background anyway.

src/assets/styles/doc-styles.scss
```diff
@@ -0,0 +1,822 @@
...
+    button, a {
+      color: inherit;
+      @apply transition-colors;
+
+      &:hover {
+        color: lighten(#e7dacb, 10%);
+      }
+    }
+
+    svg {
```
mischnic, line 372:
```suggestion
    a svg {
```

Changing this should fix the dark mode switcher color.

svgo removes `fill="black"` (which seems to be the default), but the `svg` also has `fill: currentColor`, which overrides the previously implicit `fill="black"`

mischnic:
The header color fades in if the loading is slow (uncached 3g in this case). 

https://user-images.githubusercontent.com/4586894/133938494-f285db71-59c4-4a78-90bc-bd713115dd0c.mov

<br>

And unrelated to that, there's now a flash of white background before the dark mode is applied on page load. That happened synchronously before

mischnic (approved):
Only remaining nits:
- the color transition in code blocks is different for some words
- inline code in headings doesn’t have a background as in paragraphs

PR #937, opened 2021-09-25
Update v1 to v2 migration guide

devongovett:
Also fixes the color of the scrollbar in dark mode.


src/getting-started/migration.md
```diff
@@ -1,96 +1,244 @@
...
+```
+
+{% endsamplefile %}
+{% endsample %}
 
-(The [TypeScript page](/languages/typescript) contains more informations - and limitations - of Parcel's TypeScript handling.)
+### Babel
 
-###### `@babel/preset-typescript` Isn't inserted Automatically into a Custom `.babelrc`.
+Like Parcel 1, Parcel 2 automatically detects `.babelrc` and other Babel config files. However, if you're only using `@babel/preset-env`, `@babel/preset-typescript`, `@babel/preset-react`, and `@babel/preset-flow`, Babel may no longer be necessary. Parcel supports all of these features automatically without a Babel config, and Parcel's default transpiler is much faster than Babel.
```
mischnic: You don't need a Babel config for Flow, but it's still using Babel...

devongovett: True, I almost omitted flow here... But it is true you don't need a config. 🤷

devongovett: Ok, I moved Flow to a separate section.

mischnic (approved):

PR #938, opened 2021-09-25
Remaining docs

devongovett:
Adds docs for the remaining languages, as well as minor updates to web extensions, node emulation, and debugging pages.


src/features/node-emulation.md
```diff
@@ -49,14 +71,16 @@ When (or more likely a dependency) importing packages such as `crypto`, `fs` or
...
+
+Calls to `fs.readFileSync` are replaced with the file's contents if the filepath is statically determinable and inside the project root.
 
-Calls to `fs.readFileSync` are replaced with the file's contents if the filepath is statically determinable and inside the [project root](/features/module-resolution/)
+- `fs.readFileSync("file", "utf8")` – the contents of the file as a string. The "utf8", "utf-8", "hex", and "base64" encodings are supported.
+- `fs.readFileSync("file")` – a [Buffer](https://nodejs.org/dist/latest-v16.x/docs/api/buffer.html) object. Note that the Buffer polyfill is quite large so this may be undesired.
 
-- `fs.readFileSync(..., "utf8")`: with the contants as string with (or any other valid encoding)
-- `fs.readFileSync(...)`: a Buffer object (e.g. `Buffer.from(....)` together with the an potentionally necessary polyfill)
+The `__dirname` and `__filename` variables can be used in the filename argument. String concatenation via the `+` operator and the `path.join` function may be used. Other functions, variables, or dynamic computations are not supported.
```
mischnic: Should we add that the file string (especially without using `__dirname`) is relative to the asset. With Node, it's relative to cwd

devongovett: Are you sure? I think it is relative to cwd in parcel too. I just tested...

mischnic: You mean the cwd of the `parcel build` invocation? Wouldn't that break caching?

with `project/x.txt` containing "abc" and `project/sub/x.txt` containing "sub"

```js
// project/index.js
const fs = require("fs");

console.log(fs.readFileSync("x.txt", "utf8"));
```
I get:
```
project/ $ node index.js
abc
project/ $ yarn parcel build index.js --no-cache
project/dist/index.js prints: abc

project/sub $ node index.js
abc
project/sub $ yarn parcel build index.js --no-cache
project/dist/index.js prints: abc
```

(Also I just noticed that entries on the CLI are relative to the project/entry root instead of the shell pwd?) It should have been `parcel build ../index.js` in the second case?

devongovett: hmm yeah looks like the rust code doesn't set a default relative path so it's whatever the cwd is I guess. And yeah it would break caching, good point. Maybe we should fail or ignore paths that aren't absolute instead? Hopefully no one is relying on that.

devongovett: Added a note about it for now, but we can actually fix that too.

mischnic (approved):

PR #941, opened 2021-09-29
Include default resolver in glob import example

chrisdothtml:
It wasn't immediately obvious why my build broke when I copied the `.parcelrc` content from this example, but including the default resolver fixed it. Updating the docs so it's more obvious and copy/paste-able

### DIFF
--- a/src/features/dependency-resolution.md
+++ b/src/features/dependency-resolution.md
@@ -176,7 +176,7 @@ Parcel supports importing multiple files at once via globs, however, since glob 
 ```json
 {
   "extends": "@parcel/config-default",
-  "resolvers": ["@parcel/resolver-glob", "..."]
+  "resolvers": ["@parcel/resolver-default", "@parcel/resolver-glob"]
 }
 ```
 


src/features/dependency-resolution.md
```diff
@@ -176,7 +176,7 @@ Parcel supports importing multiple files at once via globs, however, since glob
 ```json
 {
   "extends": "@parcel/config-default",
-  "resolvers": ["@parcel/resolver-glob", "..."]
+  "resolvers": ["@parcel/resolver-default", "@parcel/resolver-glob"]
```
devongovett, line 179: Hmm not sure this is right:

1. The glob resolver needs to run first.
2. The "..." should already run all of the resolvers from the default config.

What wasn't working for you? Did you get an error?

chrisdothtml: Ahh, when I tried it I omitted the `...` assuming it was filler. So nvm then, the example works as expected 👍 

Thanks!

PR #942, opened 2021-09-29
Fix RSS feed

devongovett:
Upgrades Parcel to the latest nightly which includes an XML transformer so that URLs in the RSS feed are rewritten properly. Also ensures that the feed URL is not hashed.


wbinnssmith (approved):

mischnic:
Is there an easy way to disable the hash anchors in the RSS feed only? It's not the end of the world though

Bildschirmfoto 2021-09-29 um 20 10 55

devongovett:
Yeah done. Added a custom PostHTML plugin for it.

PR #948, opened 2021-10-05
Fix broken links on Production page

danielmcormond:
I noticed these two links were broken clicking around on the current hosted docs.

### DIFF
--- a/src/features/cli.md
+++ b/src/features/cli.md
@@ -47,7 +47,7 @@ See [Production](/features/production/) for more details.
 
 ## Entries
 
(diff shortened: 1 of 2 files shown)


mischnic:
Thanks! While you're at it, could you also fix the link on this line? (`/features/targes/#entries` instead of `/features/targets/#entries`)
https://github.com/parcel-bundler/website/blob/2e7c533ff2ff7d8914725448870c87e13b07fda6/src/features/cli.md#L50

devongovett (approved):

PR #950, opened 2021-10-07
fix: Algolia search filter on version

DeMoorJasper:
Add a facet filter on version to only show results of v2 docs in v2.

Related config change https://github.com/algolia/docsearch-configs/pull/4708

### DIFF
--- a/src/assets/scripts/search.js
+++ b/src/assets/scripts/search.js
@@ -1,13 +1,13 @@
 // When the search input is focused, add a class to the header.
 // This is used to hide some other elements and expand the search bar.
-let search = document.querySelector('#search-input');
-let header = document.querySelector('.parcel-docs-header');
+let search = document.querySelector("#search-input");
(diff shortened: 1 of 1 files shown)


src/assets/scripts/search.js
```diff
@@ -1,26 +1,29 @@
...
 };
 
 docsearch({
   apiKey: "8b6be780425a72d1a1683abea2636778",
   indexName: "parceljs",
   inputSelector: "#search-input",
   algoliaOptions: {
-    facetFilters: ["lang:v2"],
-  }
+    facetFilters: ["lang:en", "version:v2"],
```
DeMoorJasper, line 18: This is the real change, ensuring we only show v2 docs and filter on actual language instead of the v2 hack.

We could also update this for v1, but worst case scenario they just end up on the v2 docs which is probably fine. Might even nudge people towards using v2 :D

devongovett (approved):

PR #975, opened 2021-10-28
vue: add type="module" for v2

nim65s:
Hi,

This example wasn't working for parcel v2.
We just need to follow https://parceljs.org/getting-started/migration/#%3Cscript-type%3D%22module%22%3E


mischnic (approved):
Thanks!

PR #981, opened 2021-11-23
Plugin browser

mischnic:
Integration questions:
- [x] Should this be integrated somewhere else (e.g. as a docs page with the search as the content)
- [x] Open for design suggestions on the actual search, it's not really consistent with the existing pages

Search Data:
- [ ] I noticed when adding `restrictSearchableAttributes` was that it changes the order to first list all non-org packages and then the scoped ones even if the scoped ones are more popular/newer


thewilkybarkid:
This will be really useful, thanks.

I've noticed that at least one result is a 404 (`@hellofresh/parcel-transformer-md`).

Is it possible to search by all types? It might be more useful to start with that, as it's not always obvious which type something is.

mischnic:
> Is it possible to search by all types? It might be more useful to start with that, as it's not always obvious which type something is.

Not sure if Algolia allows something like "parcel-transformer-* | parcel-reporter-* | ..." was a search term (wildcards and alternatives)

> I've noticed that at least one result is a 404 (@hellofresh/parcel-transformer-md).

I would guess that this is either a private package that is somehow still indexed by Algolia, or it was recently deleted.

devongovett (approved):

PR #982, opened 2021-11-28
Design updates for plugin browser

devongovett:
@mischnic wdyt?

Only did dark mode so far and didn't touch mobile. Didn't change too much, just tried to integrate with the rest of the site. Ideally it might go through eleventy first so it has the layout? Not sure where in the left nav to put the item. Maybe underneath plugins in features?


mischnic:
Looks great!

If we didn't have the link from the landing page, I would say that it's too well hidden in the sidebar? Maybe add put it in the docs/blob switcher?

I noticed that the color transition is slightly different:

https://user-images.githubusercontent.com/4586894/143771435-c6d48020-e8aa-46b6-b981-cdf8dca67d3e.mov

And personally, I also find it odd when the page content is in English but then some elements (the relative publish date) are in another language 🤷

devongovett:
> If we didn't have the link from the landing page, I would say that it's too well hidden in the sidebar? Maybe add put it in the docs/blob switcher?

Hmm I was actually thinking of removing it from the landing page just to keep things simple. If it's in the sidebar then doesn't really make sense there. One problem with the docs/blog switcher is that it's already pretty cramped up there on mobile.

In addition to the sidebar, I was also thinking of adding links from each plugin type section in "Plugins" to eg "Find Transformer plugins" or something.

mischnic:
Yeah, maybe in "getting started" or "features" in the sidebar

devongovett:
Ok, integrated it into the sidebar, and added links from each plugin to search as well. Also updated the light mode and mobile UI.

src/_includes/darkmode.js
```diff
@@ -19,3 +19,5 @@ let isDarkMode = JSON.parse(sessionStorage.getItem('isDarkMode'));
 if (isDarkMode == null) isDarkMode = isSystemDarkMode;
 
 updateMode(isDarkMode);
+globalThis.isDarkMode = isDarkMode;
+globalThis.updateMode = updateMode;
```
mischnic: I think this isn't used anymore?

```suggestion
```

mischnic (approved):
😍

PR #987, opened 2021-12-14
Use @parcel/css for minification and auto prefixing

devongovett:
Before:

```
bash-3.2$ find dist1 -name '*.css' | xargs ls -lh
-rw-r--r--@ 1 devongovett  staff    76K Dec 12 12:58 dist1/_site/404.e1ee3572.css
-rw-r--r--@ 1 devongovett  staff    54K Dec 12 12:58 dist1/src/home/index.b6e5c8a3.css
```

After:

```
bash-3.2$ find dist -name '*.css' | xargs ls -lh
-rw-r--r--  1 devongovett  staff    72K Dec 14 11:03 dist/_site/404.f7244d04.css
-rw-r--r--  1 devongovett  staff    53K Dec 14 11:03 dist/src/home/index.5757420a.css
```

### DIFF
--- a/api/bundler.html
+++ b/api/bundler.html
@@ -28,7 +28,6 @@
   div.api div.inline-method p:first-child {
     margin-top: 0;
   }
-
   div.api div.inline-method ul {
     margin: 0;
   }
--- a/.parcelrc
+++ b/.parcelrc
@@ -1,5 +1,8 @@
 {
   "extends": "@parcel/config-default",
   "resolvers": ["...", "parcel-resolver-website"],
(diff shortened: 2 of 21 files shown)


api/bundler.html
```diff
@@ -28,7 +28,6 @@
   div.api div.inline-method p:first-child {
     margin-top: 0;
   }
-
```
devongovett, line 31: This newline generated invalid css because it got embedded into markdown which treated it as a new paragraph 😂

mischnic (approved):

PR #1000, opened 2022-02-08
Add npm option to autoprefixer install under postcss section

Xaviju:
For some beginners, for instance students, npm is the default option.


devongovett:
If we do this we should apply it consistently. I believe most examples across all pages use yarn exclusively at the moment. Perhaps it would be better to have a component where you can switch between them and it remembers between pages so we don't have to write both all the time.

Xaviju:
You are right, I thought this was a missing piece, since I've seen other places in the docs where it is fixed already. I am not sure how to create the component, but I'll check the code structure and try to PR it. 
Anyway, if you think this is OK for now, it will be useful for me if merged.

Thanks

PR #1004, opened 2022-02-27
Use `devDependencies` instead of `dependencies`

kidonng:
`devDependencies` is used later in the same document:

https://github.com/parcel-bundler/website/blob/97792a0c81b5b4def7e2d31651cd2ce344bd9b1a/src/getting-started/migration.md?plain=1#L182-L184

### DIFF
--- a/src/getting-started/migration.md
+++ b/src/getting-started/migration.md
@@ -33,7 +33,7 @@ The first thing to note when upgrading from Parcel 1 to Parcel 2 is that the npm
 
 ```json/2
 {
-  "dependencies": {
+  "devDependencies": {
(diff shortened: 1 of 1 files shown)


mischnic:
Can you change it here as well?
https://github.com/kidonng/website/blob/4b0b9a015427a58670b2116a3a8b0214b2d4eacf/src/getting-started/migration.md?plain=1#L25

kidonng:
That's intentional to some extent, since some people _may_ mistakenly listed Parcel in `dependencies`. Consistency matter so I don't have an issue changing it 🤷

mischnic (approved):
Thanks

PR #1007, opened 2022-03-02
fix Visual Studio Code debug instructions

LucaBlackDragon:
Added a second `sourceMapPathOverrides` value to match current output of Parcel (see [this post on Reddit](https://www.reddit.com/r/learnjavascript/comments/sfowch/comment/hurc44w/?utm_source=share&utm_medium=web2x&context=3) for further reference).
Tested with Parcel 2.3.2 and Visual Studio Code 1.64.2


devongovett (approved):
I think that makes sense? @mischnic @DeMoorJasper?

mischnic:
I think so too. And if it's not always correct, then it's at least better than before

PR #1014, opened 2022-03-22
Update web extensions documentation

### LINKED ISSUE #1021: Web-ext: HMR mode doesn't mention, from which folder to load the extn from
https://parceljs.org/recipes/web-extension/#hmr
After I run `npm start` from which folder should I load the extn from into Chrome?

101arrowz:
Adds updated documentation on web extension MV3 support (ref https://github.com/parcel-bundler/parcel/pull/7050)

Closes #1021

### DIFF
--- a/src/recipes/web-extension.md
+++ b/src/recipes/web-extension.md
@@ -51,7 +51,7 @@ For the best developer experience, use `--host localhost` for development builds
 ```json
 {
   "scripts": {
-    "start": "parcel watch src/manifest.json --host localhost--config @parcel/config-webextension",
+    "start": "parcel watch src/manifest.json --host localhost --config @parcel/config-webextension",
     "build": "parcel build src/manifest.json --config @parcel/config-webextension"
   }
 }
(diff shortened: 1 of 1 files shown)


devongovett (approved):

src/recipes/web-extension.md
```diff
@@ -41,66 +64,80 @@ You can also create a `.parcelrc` file in your project extending `@parcel/config
 
 ## HMR
 
-For the best experience, you'll also want to enable a few options:
+Due to [restrictions on Content Security Policy](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#content-security-policy) in MV3, HMR is not supported, but updating your code will cause the extension to reload. For MV2, HMR is fully supported by default. Reloading pages with content scripts will reload the extension in both versions.
 
-- `sourceMap#inline` and `sourceMap#inlineSources`: Source maps don't work unless inlined in web extensions
```
fregante, line 46: Was this removed by mistake? I think it's still required, at least in MV2, but I haven't tested. It could only work if the new config does this automatically, in which case it should still probably be documented.

- Re: https://github.com/fregante/browser-extension-template/pull/83

101arrowz: Yes, this is automatically done for all web extensions now. You're right, this should probably be documented.

PR #1019, opened 2022-04-07
fix spelling of visualiser npm package

also:
The npm package is named `@parcel/reporter-sourcemap-visualiser`, but the usage instructions used a `z` instead of `s` in <code>visuali<strong>s</strong>er</code>.


mischnic (approved):
Thanks!

PR #1033, opened 2022-06-08
docs(cli): change --dist-dir param to --out-dir

damienbutt:
The --dist-dir parameter no longer works. As per the CLI help, the correct parameter is now --out-dir.


mischnic:
No, this is correct for Parcel 2. You must be using Parcel 1, try running `parcel --version`

damienbutt:
Ah. OK, my mistake. I will check that.

Thanks,

D

PR #1038, opened 2022-06-27
add section about compiling multiple elm modules into the same output…

ChristophP:
… file

Related to this PR https://github.com/parcel-bundler/parcel/pull/8076#pullrequestreview-1019510045

### DIFF
--- a/src/languages/elm.md
+++ b/src/languages/elm.md
@@ -75,6 +75,39 @@ view model =
 
 {% endsample %}
 
+## Compiling multiple files into a single JS output
+
+You can use the `with` query param to compile multiple Elm sources into the same bundle. This can help you keep the bundle size smaller, because things like the runtime and common code are shared.
+
+{% sample null, "column" %}
+{% samplefile "index.js" %}
+
+```js
+import { Elm } from "./Main.elm?with=./MainB.elm&with=./MainC.elm";
+
+Elm.Main.init({ node: document.getElementById("root") });
+Elm.MainB.init({ node: document.getElementById("rootB") });
+Elm.MainC.init({ node: document.getElementById("rootC") });
+```
+
+{% endsamplefile %}
+{% endsample %}
+
+This will do the equivalent of this command:
+
+```
+elm make Main.elm MainB.elm MainC.elm
+```
+
+The `with` param can be used multiple times to include multiple extra Elm programs.
+
+Beware of 2 things:
+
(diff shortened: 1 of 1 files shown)


src/languages/elm.md
```diff
@@ -75,6 +75,39 @@ view model =
...
+```
+elm make Main.elm MainB.elm MainC.elm
+```
+
+The `with` param can be used multiple times to include multiple extra Elm programs.
+
+Beware of 2 things:
+
+1. **Path base:** The paths given in the `with` param values are relative to the directory of the first file in the `import` statement (in this case `Main.elm`), NOT relative to the JS file that contains the `import` statement.
+2. **Unintentional Duplication:** If you import an import line with `with` params in multiple JS files but the query string is not exactly the same, parcel will treat it as a different asset and duplicate the content.
```
mischnic:
```suggestion
2. **Unintentional Duplication:** Multiple imports that effectively specify the same Elm files but in a different order (or differ regarding leading `./`, etc.) are treated as different assets (and will therefore be duplicated)
```

I think this is clearer

src/languages/elm.md
```diff
@@ -75,6 +75,39 @@ view model =
...
+
+To avoid those pitfalls, it's recommended to use [this extra resolver package](https://www.npmjs.com/package/parcel-resolver-elm-bundle), when making heavy use of with params (i.e. importing the bundles in more than 1 JS file, using the `with` more than 2 times)
```
mischnic:
```suggestion
To avoid those pitfalls when making heavy use of `with` params (i.e. importing some combination in more than one place), it's recommended to use something like [this third-party resolver package](https://www.npmjs.com/package/parcel-resolver-elm-bundle) which allows specifying some shorthands for commonly used Elm file combinations.
```

And it would be great if that link at least showed how the `extraSources` is supposed to work (= the docs for that package), even if it's not (fully) implemented yet.

mischnic:
And I think we should also add that having these two imports in the project will still effectively duplicate `MainB.elm`
- `./X.elm?with=./MainB.elm`
- `./Y.elm?with=./MainB.elm`

(This might be clear for people used to `elm make` but not really when coming from ESM modules)

src/languages/elm.md
ChristophP: Yeah, much clearer. Thanks

src/languages/elm.md
ChristophP: I will update that library in the next few days.

ChristophP: @mischnic I updated the readme and the implementation of the package here https://www.npmjs.com/package/parcel-resolver-elm-bundle

devongovett (approved):

PR #1039, opened 2022-07-13
Update getting_started.md

Tim-a-tree:
with the update to parcel v2, the cli for installing parcel changed to :
yarn add --dev parcel
npm install --save-dev parcel

however, korean documentation still uses cli for parcel v1:
npm install -g parcel-bundler
yarn global add parcel-bundler

so, I propose these changes to follow up with parcel v2.


mischnic:
The master branch contains the old changes for Parcel 1 (what you're editing). The v2 branch has the current docs. So this is correct

PR #1051, opened 2022-10-29
docs: fix TypeScript resolution path in `module_resolution.md`

remarkablemark:
The TypeScript resolution was giving me the error:

```
error TS2307: Cannot find module '~/src/...' or its corresponding type declarations.
```

See working example: https://github.com/remarkablemark/parcel-typescript-path-alias-demo/blob/master/tsconfig.json


devongovett:
These changes are for the v1 site, which is no longer maintained.

PR #1052, opened 2022-11-01
Document changing the webextension manifest name

mischnic:


src/recipes/web-extension.md
```diff
@@ -48,6 +49,12 @@ To build your extension, run Parcel using your `manifest.json` as an entry, and
 parcel build manifest.json --config @parcel/config-webextension
 ```
 
+{% warning %}
+
+With the default Web Extension config, the manifest has to be called `manifest.json` (and cannot be just any file with a `json` extension).
+
+{% endwarning %}
```
mischnic, line 56: This is the first actual change.

src/recipes/web-extension.md
```diff
@@ -62,6 +69,26 @@ You can also create a `.parcelrc` file in your project extending `@parcel/config
...
+    "some-other-manifest.json": ["@parcel/transformer-webextension"]
+  },
+  "packagers": {
+    "some-other-manifest.json": "@parcel/packager-webextension"
+  }
+}
+```
+
+{% endsamplefile %}
+{% endsample %}
```
mischnic, line 90: This is the other change, everything else was caused by Prettier for some reason.

akash07k:
Seems fine for now, however, actual flexibility of having any file name for the manifest will be very much helpful in future.

thebriando (approved):

fregante (approved):

mischnic:
Turns out this actually doesn't work yet because of this hardcoded name check:
https://github.com/parcel-bundler/parcel/blob/c3bbe0a6160186f496ca2f9e9bead9376c0522f1/packages/runtimes/webextension/src/WebExtensionRuntime.js#L18
We'll have to change that to just rely on some asset meta property...

akash07k:
Exactly, that will be great.

kuroppe1819:
If you stop hard-coding and rely on meta property, the build will go through. However, since the output file name remains the same, it cannot be loaded as a browser extension. I want to convert the file name to `manifest.json` when outputting, but I can't think of a good way to do it...😞. Do you have any good ideas?

sergey-shpak:
@kuroppe1819 using namer? https://parceljs.org/plugin-system/namer

PR #1053, opened 2022-11-01
Upgrade deps

mischnic:
~Blocked by https://github.com/parcel-bundler/lightningcss/issues/325~

The big breaking changes were:
- Newer (more strict) version of lightningcss
- Tailwind 3


src/assets/styles/doc-styles.scss
```diff
@@ -739,13 +739,15 @@ body {
...
     }
 
+    /*
     @page :left {
       margin: 15mm 20mm 15mm 10mm;
     }
 
     @page :right {
       margin: 15mm 10mm 15mm 20mm;
     }
```
mischnic, line 749: Not sure what that's even supposed to mean, compiled to css:
```css
@media print {
  @page :left {
    .main-content {
      margin: 15mm 20mm 15mm 10mm;
    }
  }
}
```

.parcelrc
```diff
@@ -1,11 +1,11 @@
 {
   "extends": "@parcel/config-default",
   "resolvers": ["...", "parcel-resolver-website"],
-  "namers": ["parcel-namer-website", "..."],
   "transformers": {
-    "*.css": ["@parcel/transformer-postcss", "@parcel/transformer-css-experimental"]
+    "*.{js,mjs,jsm,jsx,es6,cjs,ts,tsx}": [
+      "@parcel/transformer-js",
+      "@parcel/transformer-react-refresh-wrap"
+    ]
```
devongovett, line 8: Why did we need to override the js pipeline?

src/assets/styles/doc-styles.scss
devongovett: Theoretically if you printed the website, and made a two page spread, the left page would have different margin from the right page. But idk where this came from. What was broken with it?

.parcelrc
mischnic: [generate-api-docs](https://github.com/parcel-bundler/website/tree/v2/generate-api-docs) is written using flow, so the .flowconfig caused every file to be transpiled with Babel.

devongovett (approved):

PR #1055, opened 2022-11-05
Blog post for v2.8

devongovett:


src/blog/v2-8-0/v2-8-0.md
```diff
@@ -0,0 +1,56 @@
...
+
+The team at Atlassian has contributed a new bundling algorithm that solves these problems, significantly improving both build time and runtime performance. It takes a different approach than the previous implementation: rather than initially placing all assets into bundles based on manual code split points (e.g. dynamic import) and then removing duplication afterward, it starts with a graph containing no duplication (each asset is in only one bundle) and then combines bundles as needed to meet constraints like parallel request limits and minimum bundle size requirements. It also reduces time complexity by pre-computing more information and removing nested graph traversals.
+
+This results in both smaller bundles and much faster builds. For a very large project, overall build time was reduced from over 25 minutes to 9 minutes (**2.7x faster**). The total bundle size for the whole project went from 952 MB to 370 MB (**2.5x smaller**).
```
devongovett: @AGawrys I'd like to give some estimate of the size of the project. Do you have a count of the number of assets/bundles that I could share?

Also, any comparison with the webpack build time (if favorable 😉)?

src/blog/v2-8-0/v2-8-0.md
```diff
@@ -0,0 +1,56 @@
...
+
+This release also includes improvements to tree shaking with code splitting. Previously, if you used a large library containing an index file with many re-exports, Parcel would always place all of the used exports in the same bundle. For example, if you used different components from a component library on different pages or code split points (e.g dynamic import), all used components across your whole project would be placed into the entry bundles. This could mean loading much more JavaScript than necessary on initial page load.
+
+Parcel now rewrites dependencies to point to their final destinations, following all re-exports found along the way. This means re-exports now have no effect on code splitting, and only used exports within each bundle are loaded. In many applications, this means the entry bundle will become much smaller since more code will only be loaded when needed.
+
+Results will depend on how many re-exports you rely on and your code splitting setup, but we have seen impressive results in several applications so far. One very large app saw over **40% smaller JS entry bundles**, while another saw a 25% smaller JS entry and 50% smaller CSS.
```
devongovett: 40% was the latest number I found on https://github.com/parcel-bundler/parcel/pull/8432#issuecomment-1243070394 but lmk if there is an updated one @AGawrys @marcins

src/blog/v2-8-0/v2-8-0.md
marcins: I can get you some of the current numbers tomorrow, though that won't be the same as when we first implemented the change (so the relative improvement might not be accurate).

src/blog/v2-8-0/v2-8-0.md
```diff
@@ -0,0 +1,64 @@
...
+- Parcel is 34% faster than Turbopack, and 74% faster than Vite at updating the leaf component.
+
+See [this repo](https://github.com/devongovett/parcel-vs-vite-vs-turbopack-hmr) for full benchmark source code and results.
+
+## Tree shaking improvements
+
+This release also includes improvements to tree shaking with code splitting. Previously, if you used a large library containing an index file with many re-exports, Parcel would always place all of the used exports in the same bundle. For example, if you used different components from a component library on different pages or code split points (e.g dynamic import), all used components across your whole project would be placed into the entry bundles. This could mean loading much more JavaScript than necessary on initial page load.
+
+Parcel now rewrites dependencies to point to their final destinations, following all re-exports found along the way. This means re-exports now have no effect on code splitting, and only used exports within each bundle are loaded. In many applications, this means the entry bundle will become much smaller since more code will only be loaded when needed.
```
mischnic: > This means re-exports now have no effect on code splitting

Should we mention that this only applies to reexports in assets marked as side-effect-free?

PR #1058, opened 2022-11-16
Update node-emulation.md

Richard87:
Added note about dotenv package must be installed to parse .env-files

### DIFF
--- a/src/features/node-emulation.md
+++ b/src/features/node-emulation.md
@@ -35,7 +35,7 @@ The `NODE_ENV` environment variable is automatically set by Parcel depending on 
 
 ### `.env` files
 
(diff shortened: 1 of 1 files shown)


mischnic:
This shouldn't be necessary. What problem did you run into without adding that dependency?

Richard87:
The .env file was not parced without the dependency (tried Parcel 2.6, 2.7 and 2.8) until I added dotenv package. 

But after removing the package again it still works, so not sure whats going on now. But had the same symptoms in our "work project", couldnt get .env files to load. After installing a simple parcel hello-world page it still didn't work, adding dotenv solved the issue.

But removing the package now didn't change anything, and the .env file is still being parsed #confused

devongovett:
Maybe your package lock file got into a weird state or something.

PR #1067, opened 2022-12-02
Remove reference to nonfunctional third-party package

benwilliamsBJSS:
I get  `@parcel/package-manager: Could not find module "parcel-resolver-elm-bundle", but it was listed in package.json. Run your package manager first.` whenever attempting to utilise the package recommended.

Removing so future developers don't lose as much hair as I did.


mischnic:
@ChristophP Could you please:
- Create a github repo for that npm package and
- also fix the package.json `main` field which currently points to a non-existent index.js where it should be src/resolver.js

ChristophP:
@benwilliamsBJSS sorry if this costs you time. This got buried under some other todos.

@mischnic I published a repo here https://github.com/ChristophP/parcel-resolver-elm-bundle . Also I cleaned up some things in the package, fixed the entrypoint and published a new version. This should be set now.

ChristophP:
Still found one bug in it and also ran into this https://github.com/parcel-bundler/parcel/issues/7107 .
Now I published version `1.0.2` which worked for me and should work for hopefully everyone.

PR #1073, opened 2023-02-06
Package is called source-map, not source-maps

th3james:
Apologies if I've misunderstood this, but I believe the package name referenced is incorrect, and should be https://www.npmjs.com/package/@parcel/source-map


mischnic (approved):
You're right, thanks!

PR #1074, opened 2023-02-25
fix: remove duplicate word

cuvar:
I found a duplicated word in the documentation. Just a small pr for removing it


mischnic (approved):
Thanks!

PR #1080, opened 2023-04-18
Glob: Improve example snippet

apepper:
Without "@parcel/resolver-default" other resolves such as src/index.html will not work.


mischnic:
No, if you write `"resolvers": ["@parcel/resolver-glob", "..."]` (that is: literally, with `"..."`), then it will still use the resolvers of the extended config.

PR #1081, opened 2023-04-20
Update getting_started.md

Umka-Ness:
Do not use npm install -g parcel-bundler because installation is not possible, only npm install -g parcel works, I struggled for several hours until I read an article where they wrote that it no longer works


mischnic:
These are the docs for Parcel 1, where `parcel-bundler` was correct.

PR #1090, opened 2023-09-07
fix: stretched out sidebar icons

Birdie0:
Non-square icons on the sidebar are stretched out, adding `object-fit: contain` allows to keep their original aspect ratio.

| before | after |
|---|---|
|23-09-07_14-09-16_ErW65|23-09-07_14-09-07_firefox_RntXo|


mischnic (approved):
Thanks!

PR #1091, opened 2023-09-13
Update webapp.md

kswdgha:
Add viewport tag. Couldn't notice why css media didn't work for a while.


mischnic:
This page is an example of how to get started with Parcel. I can't list everything that's needed for a modern webapp

PR #1099, opened 2023-12-14
Add Manual Shared Bundle Blurb to Code splitting doc

AGawrys:
Adds documentation on how to use the `manualSharedBundles` config option for the `DefaultBundler`. Includes some common reasons why a project may benefit from MSB (manual Shared Bundles), as well as a basic example of how to add config to your `package.json`

This PR should be merged once the [Source Bundles Property PR](https://github.com/parcel-bundler/parcel/pull/9294) is merged. That PR renames `parentAsset` to `root`, the property that specifies what file should be targeted by the `assets` glob. 

It also adds a `loadedBy` optional config, which allows users to specify bundleRoots that will use this MSB (manual shared bundle)
## Notes
Note that "display rich diff" in Github does not support all formatting used, here's a screen grab of what the page will look like: 
 
experimental_msb_blurb


devongovett:
Updated to remove loadedBy for now since that PR isn't merged yet, and re-worded a bit for clarity.

devongovett (approved):

PR #1103, opened 2024-02-11
Add docs for macros

devongovett:


src/features/macros.md
```diff
@@ -0,0 +1,355 @@
...
+<div className={css('color: red; &:hover { color: green }')}>
+  Hello!
+</div>
+```
+
+{% endsamplefile %}
+{% samplefile "css.ts" %}
+
+```ts/6-10
+import type {MacroContext} from '???';
```
devongovett: Where should we put this type?

mischnic:
👍 Looks good

PR #1104, opened 2024-02-27
Add REPL link to website header

devongovett:


src/assets/styles/doc-styles.scss
```diff
@@ -347,6 +347,15 @@ body {
...
     }
+
+    @media (width < 430px) {
+      a {
+        display: none;
+      }
+      #color-scheme-toggle {
+        margin-left: 0;
+      }
+    }
```
devongovett, line 358: Hides the GitHub link on small mobile screens since it is also in the hamburger menu

mischnic (approved):

PR #1108, opened 2024-04-28
plugins.md: replace reference to old `link:` protocol with `file:`

ynx0:
The `link:` protocol was renamed to `file:` so this documentation is outdated. (source: [this SO post](https://stackoverflow.com/a/74679690))

This PR updates the docs to reflect that.


ynx0:
Looks like the section header in the search still shows up with the old value. Not sure how to fix that.

mischnic:
No, for Yarn, `link` and `file` both exist and do different things.

ynx0:
Oh I see. I think it would be useful to at least add a note for npm users on what to do. I can make a separate PR for that.

